PR #2, opened 2017-11-04
Add SSH extension

benedictdudel:
When the extension is installed one is able to trigger it by typing `ssh`. When no query is given Albert will show hostnames from `~/.ssh/config` and `/etc/hosts`. When there is a query the extension will search through all available hostnames and will suggest them. Another action is that you can `ping` a host as well.

Currently I don't know which icon to show. Maybe you got some idea?

### DIFF
--- a/SSH.py
+++ b/SSH.py
@@ -1,0 +1,112 @@
+""" SSH """
+
+import subprocess
+import getpass
+from albertv0 import *
+from shutil import which
+from pathlib import Path
+
+__iid__ = "PythonInterface/v0.1"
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:
Actually there is a native ssh extension? Why do you think this is necessary?

benedictdudel:
Well, I totally overlook that... Since the native ssh extension covers all of my use cases for the extension I close the pull request.

ManuelSchneid3r:
I am sorry abou that. However thank you for yout efforts.

PR #4, opened 2017-11-04
Add Kill extension

benedictdudel:
The kill extension lists all running processes and the user can kill a running process by choosing a process from the list.

### DIFF
--- a/Kill.py
+++ b/Kill.py
@@ -1,0 +1,71 @@
+""" Kill Process Extension """
+
+import os
+import getpass
+from albertv0 import *
+from subprocess import Popen, PIPE
+
+__iid__ = "PythonInterface/v0.1"
+__prettyname__ = "Kill Process"
+__version__ = "1.0"
+__trigger__ = "kill "
+__author__ = "Benedict Dudel"
+__dependencies__ = []
+
+iconPath = iconLookup('process-stop')
+
+def handleQuery(query):
+    if query.isTriggered:
+        results = []
+        for process in getProcesses(query.string.strip()):
+            results.append(
+                Item(
+                    id=process['id'],
+                    icon=iconPath,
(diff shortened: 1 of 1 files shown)


Kill.py
```diff
@@ -0,0 +1,59 @@
...
+__dependencies__ = []
+
+def handleQuery(query):
+    if query.isTriggered:
+        results = []
+        for process in getProcesses(query.string.strip()):
+            results.append(
+                Item(
+                    id=process['id'],
+                    icon=iconLookup(process['name']),
```
ManuelSchneid3r: Maybe [*process-stop*](https://standards.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html#names) makes sense. What do you think? Looks somewhat like this on my system 

image

Kill.py
```diff
@@ -0,0 +1,59 @@
...
+                    text=process['name'],
+                    subtext="CPU: %s %%" % process['cpu'],
+                    completion="",
+                    actions=[
+                        ProcAction("Kill", ["kill", "-SIGKILL", process['id']]),
```
ManuelSchneid3r: I think sigterm is a better default and as alternative action sigkill

benedictdudel: Sounds good :)

Kill.py
```diff
@@ -0,0 +1,71 @@
...
+
+import os
+import getpass
+from albertv0 import *
+from subprocess import Popen, PIPE
+
+__iid__ = "PythonInterface/v0.1"
+__prettyname__ = "Kill Process"
+__version__ = "1.0"
+__trigger__ = "kill"
```
ManuelSchneid3r: Please use space after triggers, otherwise "Killing Hasselhoff" would trigger this extension

PR #1, opened 2017-11-02
Add CopyQ extension for clipboard management

benedictdudel:
I migrated the old CopyQ extension and added the following features:

- search by value or with copyq entry number
- remove action to remove a specific clipboard item

Let me know what you think!

PR #12, opened 2017-11-23
Adding clipboard support using copyq

raghuramos1987:
Is there a way I can get the "Amount of results" setting in python? I could use that to limit the number of items in the for loop to just get the right number.

### DIFF
--- a/Clip.py
+++ b/Clip.py
@@ -1,0 +1,35 @@
+from albertv0 import *
+from subprocess import call, check_output
+from shutil import which
+import random, string
+
+__iid__ = "PythonInterface/v0.1"
+__prettyname__ = "Clipboard"
+__version__ = "1.0"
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:
Stooop  :D we have 4 versions of CQ now :D

raghuramos1987:
Oh! I did not know that. We can close this one then. Is there documentation on how to use the others?

ManuelSchneid3r:
There is not much to document type `cq` to see the entries and and an optional filter to filter them

benedictdudel:
@raghuramos1987 If you would like to contribute further please have a look at the following [pull request](https://github.com/albertlauncher/python/pull/1) and give some feedback if you like to :)

raghuramos1987:
will do @benedictdudel !

PR #16, opened 2017-11-25
[scrot] Improve saving of screenshot and extension name

benedictdudel:
Make use of the environment variable `XDG_PICTURES_DIR` when creating
screenshot. If those variable is not set we save the screenshot in the
temp directory.

Additionally we need to rename the extension name to give users a hint
which don't know `scrot`.

@ManuelSchneid3r Do you know a way to delay the execution of the command? `scrot` doesn't support a value below 1 second.

@schoettl Thanks for your review. I tried to implement you suggestions in the old pull request. Would like to see your feedback :)

### DIFF
--- a/Scrot.py
+++ b/Scrot.py
@@ -5,13 +5,16 @@ screen, an specific area or the current active window.
 When the screenshot was made you will hear a sound wich indicates that the
 screenshot was taken successfully.
 
-Screenshots will be saved in your home directory by default."""
+Screenshots will be saved in XDG_PICTURES_DIR or in the temp directory."""
 
 from albertv0 import *
 from shutil import which
+import subprocess
+import tempfile
+import os
 
 __iid__ = "PythonInterface/v0.1"
-__prettyname__ = "scrot"
+__prettyname__ = "scrot screenshot utility"
 __version__ = "1.0"
 __trigger__ = "scrot "
 __author__ = "Benedict Dudel"
@@ -25,7 +28,6 @@ if which("xclip") is None:
     raise Exception("'xclip' is not in $PATH.")
 
 iconPath = iconLookup("camera-photo")
-fileName = "%Y-%m-%d-%T-screenshot.png"
 
 
 def handleQuery(query):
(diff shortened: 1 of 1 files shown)


Scrot.py
```diff
@@ -25,7 +27,10 @@
     raise Exception("'xclip' is not in $PATH.")
 
 iconPath = iconLookup("camera-photo")
-fileName = "%Y-%m-%d-%T-screenshot.png"
+screenshotDirectory = os.getenv("XDG_PICTURES_DIR", tempfile.gettempdir())
```
schoettl: In my understanding, the XDG dirs are a bit more complicated to query: The environment variables are normally not defined. They would only be defined if you sourced the `~/.config/user-dirs.dirs` in a shell.

To query the directory correctly, it looks like you have to call this command: `xdg-user-dir PICTURES` - its output is the path to use. If empty, use the temp dir. See the last lines of this link for reference: https://wiki.archlinux.org/index.php/XDG_user_directories

Then the xdg package (or xdg-user-dir?) should be added as dependency.

Scrot.py
```diff
@@ -25,7 +27,10 @@
...
+fileName = os.path.join(screenshotDirectory, "%Y-%m-%d-%T-screenshot.png")
+xclipCommand = "xclip -selection c -t image/png < $f"
+scrotCommand = ["scrot", "--delay", "1", "--exec", xclipCommand, fileName]
```
schoettl: Looks very clean now :)

ManuelSchneid3r:
@benedictdudel use a FuncActio

Scrot.py
benedictdudel: ah, thanks I totally missed that..

ManuelSchneid3r:
@benedictdudel Sorry wrote from my phone and hit enter by accident GihubMobile does not allow to edit messages. Use a FuncAction and sleep before run.

benedictdudel:
@ManuelSchneid3r No problem :) just needed a hint :P

PR #17, opened 2017-11-26
Add ip address extension

benedictdudel:
This extension finds out your internal and external ip address. You just need to type `ip ` and you get two items with the found addresses.

### DIFF
--- a/Ip.py
+++ b/Ip.py
@@ -1,0 +1,58 @@
+"""The extension finds out your internal and external ip address"""
+
+from shutil import which
+from albertv0 import *
+import subprocess
+import socket
+import os
+
+
+__iid__ = "PythonInterface/v0.1"
+__prettyname__ = "IP Addresses"
+__version__ = "1.0"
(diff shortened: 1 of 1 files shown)


Ip.py
```diff
@@ -0,0 +1,58 @@
...
+    raise Exception("'curl' is not in $PATH.")
+
+iconPath = iconLookup("preferences-system-network")
+
+
+def handleQuery(query):
+    if not query.isTriggered:
+        return None
+
+    proc = subprocess.run(["curl", "ipecho.net/plain"], stdout=subprocess.PIPE)
```
ManuelSchneid3r: Please use urllib for that to avoid the dependencies. Afaik curl is not always available.

ManuelSchneid3r:
Maybe several ip providers would be cool. For the case this one has availablilizty problems

ManuelSchneid3r:
nvm it'll work :+1: Thank you

PR #18, opened 2017-11-30
Add search using gnome-dictionary

nikhilw:
Just another dictionary search extension.


ManuelSchneid3r:
Does gnome-dict have an API? It would be cool to see the results instead of starting a new search.

nikhilw:
I agree, but could not find any API. It is a little too simplistic dictionary.

PR #19, opened 2017-12-07
Added brief description, usage and example

faheel:
- Added a brief description, usage and example for the extensions.
  Eg:
  ```python
  """Convert currencies using Google Finance.
  Usage: exch <amount> <src currency> <dest currency>
  Example: exch 5 usd eur"""
  ```

- Fixed some typos.


ManuelSchneid3r:
Nice thank you :D

PR #20, opened 2017-12-20
Apt plugin added

harindu95:


Apt.py
```diff
@@ -0,0 +1,66 @@
+"""locate adapter extension
```
ManuelSchneid3r: Description is missing.

Apt.py
```diff
@@ -0,0 +1,66 @@
...
+import re
+
+__iid__ = "PythonInterface/v0.1"
+__prettyname__ = "Apt"
+__version__ = "1.0"
+__trigger__ = "apt"
+__author__ = "Harindu Dilshan"
+__dependencies__ = ['apt','gksu','pkexec']
+
+if which("apt") is None:
```
ManuelSchneid3r: maybe using apt-get is better since apt is a userfriendly frontend that may break backward compatibility

Apt.py
```diff
@@ -0,0 +1,66 @@
...
+from shutil import which
+from albertv0 import *
+import re
+
+__iid__ = "PythonInterface/v0.1"
+__prettyname__ = "Apt"
+__version__ = "1.0"
+__trigger__ = "apt"
+__author__ = "Harindu Dilshan"
+__dependencies__ = ['apt','gksu','pkexec']
```
ManuelSchneid3r: why dont you just prepend the command with sudo? sudo is way mor portable, in fact gksu is not available most of the times

ManuelSchneid3r: ah you do in line 26

Apt.py
```diff
@@ -0,0 +1,66 @@
...
+    raise Exception("'Apt' is not in $PATH.")
+
+gksu = None
+for item in __dependencies__:
+    if which(item):
+        gksu = item
+
+UpdateActions = [TermAction("Update database in Terminal", ["sudo", "apt", "update"])]
+
+if gksu is not None:
```
ManuelSchneid3r: I'd recommend to remove this. You dont want to mess with the package manager. users should always see the output of it.

ManuelSchneid3r: which would remove gksu from the dependencies. I recommend this anyway because the workflow is exactly the same but there is one additional dependency.

Apt.py
ManuelSchneid3r: pkexec is actually not needed

Apt.py
```diff
@@ -0,0 +1,66 @@
...
+    UpdateActions.insert(0,FuncAction("Update database",lambda :subprocess.Popen([gksu,"apt","update"])))
+
+iconPath = iconLookup("system-software-install")
+
+def handleQuery(query):
+    results = []
+    if query.isTriggered:
+        if len(query.string) > 2:
+            # pattern = re.compile(query.string, re.IGNORECASE)
+            shell_cmd = ["apt-cache" ,"search"]
```
ManuelSchneid3r: apt-cache is not in the deps list. I am not good with debian. If it is always installed thats fine. if not users should know it.

ManuelSchneid3r:
I dont know how to do this, but could you please finish the change request when you made your changes? Last time I did not recognize your changes since the [PR overview](https://github.com/albertlauncher/python/pulls) told me that there are still changes requested.

image

ManuelSchneid3r:
@harindu95? R u still there?

ManuelSchneid3r:
Seems to be abandoned by OP.

PR #21, opened 2017-12-20
Firefox history

harindu95:


Tyilo:
You should only add the firefox plugin in this pull request. You're also adding your apt plugin.

ManuelSchneid3r:
I am not sure if it is a good idea to add firefox twice. There is a native extension.

harindu95:
@ManuelSchneid3r  It only supports Firefox bookmarks , while it could offer more.

ManuelSchneid3r:
> @ManuelSchneid3r It only supports Firefox bookmarks , while it could offer more.

Sure but this should be part of the native firefox extension then. User experience plays a role. And well UX is not that high atm to say the least. I dont want to make it even worse.

ManuelSchneid3r:
created an issue. has to be ported into native extension.

ManuelSchneid3r:
I'am sorry to reject this @harindu95. but as said [here](https://albertlauncher.github.io/docs/contributing/)  

Do absolutely not just implement things and send a PR! Contact us / Open an issue / Comment an issue. This way nobody has to reject PR’s, nobodies PR’s will get rejected and there won’t be multiple developers working on the same issue.

FirefoxHistory.py
```diff
@@ -0,0 +1,79 @@
...
+    elif len(txt.strip()) == 0:
+        return True
+
+    return False
+
+def setupSession():
+    conn = sqlite3.connect(db_path) 
+    cursor = conn.cursor()
+    query = 'select p.url,p.title from moz_historyvisits as h, moz_places as p where p.id == h.place_id order by h.visit_date desc;'
+    query2 = 'select p.title from moz_historyvisits as h, moz_places as p where p.id == h.place_id order by h.visit_date desc;'
```
idkCpp: Does `query2` even do anything?

PR #22, opened 2018-01-17
Add spell extension

marek-mazur:


ManuelSchneid3r:
What is this keep file for?

marek-mazur:
Git doesn't track empty directories. Having a .keep file in it, makes the directory to be tracked.

marek-mazur:
@ManuelSchneid3r Can you merge it?

ManuelSchneid3r:
nobody reviewed it yet

Spell/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+
+__iid__ = "PythonInterface/v0.1"
+__prettyname__ = "Spell checker"
+__version__ = "2.0"
+__trigger__ = "spell "
+__author__ = "Marek Mazur"
+__dependencies__ = []
+
+
+dict_file = '~/.local/share/albert/org.albert.extension.python/modules/Spell/dictionaries/{language}.dict'
```
schoettl: you should better use another way to get the ~/.local/share directory. see https://askubuntu.com/questions/457047/how-can-i-get-the-xdg-default-user-directories-from-python

schoettl: is it a good idea to rely on the shared data of another extension? i'm not sure how the dictionaries go into org.albert.extension.python/... anyway?

Spell/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+    Find in dictionary
+
+    :param str language: Language
+    :param str phrase  : Phrase
+
+    :return str
+    """
+    a_dict_file = dict_file.format(language=language)
+    cmd = "grep '^{phrase}' -m {limit} {dict_file}".format(phrase=phrase, limit=limit, dict_file=a_dict_file)
```
schoettl: that's not good practice, using format to create a shell command! (shell code injection!) you should use a function from subprocess to which you pass an list, e.g. `subprocess.call(['grep', '-m', limit, '^' + escape_regex(phrase), dict_file])`

Spell/__init__.py
ManuelSchneid3r: use [this](https://albertlauncher.github.io/docs/extensions/python/#extension-interface-02) to get the path prefixes

marek-mazur: another extension? it is for this extension only. Dictonaries must prepared manually as I described here https://github.com/marek-mazur/albert-spell-extension/blob/master/README.md

marek-mazur: Hmm... But there is no README.md file in this PR. Should i add it or add instructions to dock block?

schoettl: ok, i didn't see the README.
still, i thought the org.albert.extension.python directory is for the *python* extension of albert. i thought another extension would have an org.albert.extension.otherext folder. but sorry if i'm wrong, didn't read the spec lately.

marek-mazur: Yep. This repository shouldn't be named "python". Better idea is rename it to extensions and totally deprecate old extension repository... @ManuelSchneid3r Will you consider that to prevent mistakes like this in future?

ManuelSchneid3r: Somewhere in the faar future other languages may be added. Therefore *extensions* would be ambiguous.


> i thought the org.albert.extension.python directory is for the python extension of albert.

it is. i think relative paths would be better to access resources shipped with the python module. 

persisitent user data should go to $(dataLocation())/<plugin_id>/ e.g. ~/.local/share/albert/spell/

spell is the id here. i need to unify the identifiers of the extensions, but I have absolutely no time atm. In future python extension should have an identifier too. This means spell id would be org.albert.extension.python.spell and the datalocation therefore ~/.local/share/albert/org.albert.extension.python.spell/. On the other hand I am thinking about  removing this overprecise identifiers.

ManuelSchneid3r:
@marek-mazur are you still interested in this one to get merged?

marek-mazur:
@ManuelSchneid3r yes. I will fix all bugs in this weekend.

Spell/__init__.py
marek-mazur: fixed

marek-mazur: fixed

Spell/__init__.py
marek-mazur: fixed

marek-mazur:
and?

marek-mazur:
aaand? :-P

ManuelSchneid3r:
How is this extension intended to be used? Where are the dicts from?

marek-mazur:
@ManuelSchneid3r Please look at README.md file.

ManuelSchneid3r:
@marek-mazur this is to cumbersome for ppl to use. Aint there a way to not touch the terminal, except from installing the dependencies? Why dont you just use the command directly?

ManuelSchneid3r:
somhow like this

```echo oscillatin | aspell -l EN -a 
@(#) International Ispell Version 3.1.20 (but really Aspell 0.60.6.1)
& oscillatin 4 0: oscillating, oscillation, oscillator, oscillate
```

but I also found this

marek-mazur:
It is impossible to get same results like from dumped dictionary, even with this python library.

marek-mazur:
To make it work out of box a have attached generated and compressed dictionaries (21 popular languages) (187 MB). Probably it's not good idea to keep them in github but now nobody will have any problems and it will works just after installation.

marek-mazur:
As I said before there is no possibility to have results like I want without dumping aspell dictionaries. With aspell you can only find similar words with small difference or just show misspelled words. For example:

```
echo kil | aspell -l EN -a
@(#) International Ispell Version 3.1.20 (but really Aspell 0.60.7-20110707)
& kil 21 0: Kiel, kill, kilo, kl, kiln, kilt, IL, Gil, KIA, Kim, Kip, Kit, ail, kid, kin, kip, kit, mil, nil, oil, til

echo gór | aspell -l PL -a
@(#) International Ispell Version 3.1.20 (but really Aspell 0.60.7-20110707)
+ góry

```


My goal Was to show the entry as in a real printed dictionary. Examples:
spell_en_kil
spell_pl_gor


I am just tired waiting so many months for merge simple extension that worked fine from the begining...

ManuelSchneid3r:
Okay I think I got it now. Yes, thank you for your time and patience. Extensions should be extensive in their user experience. In my eyes there is no reason why the extension should not dump the dicts on its own.  Im am on the road. Ill check aspell later. If there is a way to list the installed dicts then this could all be automated in the initialize function.

ManuelSchneid3r:
Sure the extension may be working fine for you. But like this nobody will use it. Thats not the point of sharing extensions upstream. I am thankful for every single LOC i get but somehow I have to guarantee some quality regarding the user experience.

marek-mazur:
Idea of dumping dictionaries in extension sounds good, but the problem is that it can take few minutes (!) according to dictionary size. Is it possible to create some install function (that will create dictionaries if they not exist)? It should be launched earlier than first extension usage - maybe just after enabling extension in application settings.

marek-mazur:
@ManuelSchneid3r Done. You can test it now. On my computer with three languages (en, de, pl) it takes few seconds.

marek-mazur:
156 lines of code.... and 8 months of review... for now. ;-)

ManuelSchneid3r:
> 156 lines of code.... and 8 months of review... for now. ;-)

Sure. Please be patient. We all have our RLs.

ManuelSchneid3r:
Hey @marek-mazur. I cant merge this one. Its simply to big. The repos size exploded due to the dicts. GitHub stores everything… if you manage to make the dicts be genereated at runtime everything would be fine.

Further there is a new [community repository for python extensions](https://github.com/albertlauncher/python_community) now. Please send the PR there. Sorry for the inconvenience. I am just trying to reduce the workload here and make you happy with contemporary merges. If you want to join the party (write access) ask for it in the community chats.

marek-mazur:
@ManuelSchneid3r You didn't even read this code right? Auto generating dicts was done a half year ago. On first run it will generate all installed dicts from aspell. I have removed pregenerated dicts from repository. But I don't care now...

ManuelSchneid3r:
> But I don't care now...

Relax @marek-mazur. This project is a loadshit of work and nobody is willing to help me. As in physics, keeping (man-)power constant, an increase of work implies an increase of time. Its that easy. I am not going to sacrifice myself for this project. Although I actually do. I often ask myself if I am batshit insane. Doing this stuff each an everyday while neglecting my RL.

Back to topic:

> I have removed pregenerated dicts from repository.

bildschirmfoto vom 2019-02-04 11-23-00

Thats what I see [here](https://github.com/albertlauncher/python/pull/22/files).

PR #25, opened 2018-02-04
Show which command we install aur packages with

Tyilo:


ManuelSchneid3r:
Okay I see the point but there are some things I want to improve. Ill change this. Thank you.

PR #5, opened 2017-11-05
Add Window Switcher extension

benedictdudel:
This extension can interact with a EWMH/NetWM compatible X Window Manager.

Usage:
- list and filter all windows by typing `wm <filter>`
- list and filter desktops by typing `wm desktop <filter>`
- Default action:
  - switch to window/desktop
- Alternative actions:
  - close window

Dependencies:
- wmctrl
- xprop

PR #27, opened 2018-02-05
Added Switch Application

edjoperez:
Adds functionality to toggle between windows applications

image

image

It doenst use its  trigger, so it makes albert has a more natural experience when switch between applications or new ones. Like rofi

### DIFF
--- a/SwitchApp.py
+++ b/SwitchApp.py
@@ -1,0 +1,68 @@
+# -*- coding: utf-8 -*-
+import subprocess
+from albertv0 import *
+
+#METADATA
+__iid__ = "PythonInterface/v0.1"
+__prettyname__ = "WindowSwitcher"
+__version__ = "1.0"
+__trigger__ = "!"
+__author__ = "Ed Perez"
+__dependencies__ = []
+
+def initialize():
+    pass
+
+def handleQuery(query):
+    if query.string:
+        stripped = query.string.strip()
+        results = []
+        for win in window_search(stripped):
+            results.append(Item(id=__prettyname__,
+                                icon=window_icon(win['wm_class']),
+                                text=win['wm_class'].split('.')[-1].replace('-',' '),
+                                subtext=win['wm_name'],
+                                actions=[ProcAction("Switch Window",
+                                                    ["wmctrl", '-i', '-a',win['wid']])]))
+        return results
+
+def window_list():
+    _list = []
+    out = subprocess.check_output(['wmctrl', '-l', '-x'])
+    for line in out.splitlines():
+        args = line.split(None,4)
+        wm_class = args[2].decode('utf-8')
+        if window_filter(wm_class):
+            _list.append({
+                'wid':args[0].decode('utf-8'),
+                'wm_class':wm_class,
+                'wm_name':args[4].decode('utf-8') if len(args) > 4 else ''
+            })
+    return _list
+
+def window_search(filter=''):
+    _list = []
+    for win in window_list():
+        if filter in win['wm_name'] or filter in win['wm_class']:
+            _list.append(win)
+    return _list
+
+def window_icon(filter=''):
+    icon_name = filter.split('.')[0]
+
+    if filter in 'sun-awt-X11-XFramePeer.jetbrains-pycharm':
+        icon_name='pycharm'
+
+    return iconLookup(icon_name)
+
+def window_filter(filter=''):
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:
Did you notice that there is https://github.com/albertlauncher/python/pull/5/files

edjoperez:
The only difference is that this one gives you a more fluent experience when you select the app you want to launch, because it adds the applications that are already running and the applications that you would launch

Raensul:
if i fv switching from phpstorm to chrome it is no focused window. i must click to chrome's window and then press ctrl+T for example

husimo:
Hello. It works great here under Opensuse Leap 15.0.
Just a bug : 
- `Firefox` triggers the "Firefox" window where as `firefox` doesn't work.
- `Konsole` triggers the "konsole" window, the same as `Konsole`.
So it seems the script cannot properly detect windows where their name contains a uppercase letter.
Thank you

ManuelSchneid3r:
Conversation locked until [poll](https://github.com/albertlauncher/albert/issues/123#issuecomment-413868346) finished.

SwitchApp.py
```diff
@@ -0,0 +1,68 @@
...
+        wm_class = args[2].decode('utf-8')
+        if window_filter(wm_class):
+            _list.append({
+                'wid':args[0].decode('utf-8'),
+                'wm_class':wm_class,
+                'wm_name':args[4].decode('utf-8') if len(args) > 4 else ''
+            })
+    return _list
+
+def window_search(filter=''):
```
ManuelSchneid3r: as @husimo said please make this case insensitive.

ManuelSchneid3r:
There is further the question how the heck we can get the correct icon paths. E.g. spotify has the icon name _spotify-client_. However it has to be possible because the window manager show the icons correctly.

ed-fruty:
Works for almost all my applications, but won't switch to my `PHPStorm` running instance, only suggest to open a new one instance.

monkeydri:
Does not work for me, SwitchyApp 1.2 activated and `wmctrl` installed but no existing window appear in the results.

divyanshu013:
Hey guys, how do I use this feature for switching apps?

ManuelSchneid3r:
I am not sure anymore is this extension shipped in 14.22? Otherwise you have to download and install it manually.

offzale:
@ManuelSchneid3r is this included in version 0.16.1? It is not working for me installing the rpm package in fedora 29

kuamatzin:
I just enabled the extension but it does not work, I already install wmctrl, do I need some extra step?

Mimameid:
> Hey guys, how do I use this feature for switching apps?
Bump

jason--liu:
Hey guys, Im still wondering how to trigger this feature? thanks.

ManuelSchneid3r:
> I just enabled the extension but it does not work, I already install wmctrl, do I need some extra step?

What does stdout tell?

rosaLux161:
Any updates on this?

justpeanuts:
Sorry if this is a dumb question, but is this feature currently there in albert? - switch to an application window if it is already open instead of launching another instance (without using the seperate menu item to switch to an open window using the "window-switcher.py" python extension)

Thank you

zoli:
> Sorry if this is a dumb question, but is this feature currently there in albert? - switch to an application window if it is already open instead of launching another instance (without using the seperate menu item to switch to an open window using the "window-switcher.py" python extension)
> 
> Thank you

Yes in python plugins there is a window switcher that you should enable. Then for example if you have firefox open, When you search firefox in albert it brings another entry other that switches to running instance of firefox.

tuttiamore:
The problem appears to be related to faulty destructuring of the window names into variables - the code assumes that window names always consists of two elements seperated by a `.`, when in fact window names can consists of more than two elements.

A quick hack for people interested in making this work:
Change the code in the python module on line 28 from: 
`win_instance, win_class = win.wm_class.replace(' ', '-').split('.')`
to 
`win_instance, win_class = win.wm_class.replace(' ', '-').rsplit('.',1)`

ManuelSchneid3r:

ManuelSchneid3r:
https://github.com/albertlauncher/albert/issues/123#issuecomment-413868346

This poll decided on this plugin. If anyone interested in porting it to api v2.x please send a PR.

PR #29, opened 2018-02-21
adding CopyQ Snippets plugin

turuflowers:
For those who are using CopyQ and already had a tab with Snippets, or for those who the kv it's not enough, here's an adaptation from the CopyQ Clipboard Management.

You need a tab called "snippets", where you can name your entries as you wish (yes, even with blank spaces, I think kv can't do that :smirk:), and also with the fantastic pasting after your selection.

You can also search between the name and the actual value. For example:

Key | Value
--- | ---
VPN | sudo openconnect domain.com -u user@email.com 
VPN pass | 123456 

* If you type "vpn" you get both results
* If you type "openconnect" you get the 1°
* If you type "pass" you get the 2°

### DIFF
--- a/Snippets.py
+++ b/Snippets.py
@@ -1,0 +1,86 @@
+"""CopyQ Snippets Clipboard Management"""
+
+import subprocess
+from albertv0 import *
+from shutil import which
+import json
+
+__iid__ = "PythonInterface/v0.1"
+__prettyname__ = "Snippets"
+__version__ = "1.0"
+__trigger__ = "s "
+__author__ = "Nicolás Flores"
+__dependencies__ = ["copyq"]
+
+
+if which("copyq") is None:
+    raise Exception("'copyq' is not in $PATH.")
+
+iconPath = iconLookup('copyq')
+
+copyq_script_getAll = r"""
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:
Hmm... This is somehow an abuse of copyq. Whats wrong with KV store?

ManuelSchneid3r:
@turuflowers i cant merge unless I am proven to be wrong with this beeing an edge case. When I merge this I encourage ppl to abuse cq. which I dont really want. Sure KV really needs to be overhauled, but its the only way to go. However thank you for your efforts. Keep it. Maybe there will be some sort of extension voting/download mechanism like atoms where simply all extensions are accepted.

turuflowers:
@ManuelSchneid3r the idea behind this extension is to make Albert more like Alfred. Alfred have great features, like the copy history and the snippets. With the CopyQ extension, Albert gets the copy history. I think that the idea of KV was to have the snippets, but the way its working now its far away from the way Alfred does, and thats why I made this, to make it work the same as Alfred.
Maybe my error was to not change KV, but I dont have enought knoweldge to do that.

PR #33, opened 2018-05-01
Changed from Google Finance to Yahoo Search to get exchange rates

hulloanson:
Google Finance url is no longer accessible.
To avoid complexities of setting headers, I used Yahoo Search instead of Google Search to replace this feature.


ManuelSchneid3r:
image

ManuelSchneid3r:
Okay I rewrote the ext. the basis is now the data form the european central bank.

PR #34, opened 2018-05-02
GoogleTranslate & MultiGoogleTranslate: Fixed TypeError when passing HTTPResponse to json lib for decoding

hulloanson:
There are 2 fixes:
1. `TypeError` when parsing response
`json.load()` raised `TypeError` when parsing responses from Google Translate. (Happened on 3.5.3)

Made change so that bytes from `response.read()` are explicitly converted to `str` and passed to `json.loads()` instead of `json.load()`.

2. Erroneous Japanese language code 
`ja` is the correct one.

Edit: Added commit fixing Japanese language code.


ManuelSchneid3r:
looks sane, ty.

PR #35, opened 2018-05-02
Currency add xe

g-walsh:
I also saw the problem with Google Finance but investigated xe.com as a provider. Both XE and Yahoo work so I added the ability to list providers, by default it uses Yahoo but if that produces an exception (like Google Finance does) then it will fallback to xe.com.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 __pycache__
-
+.mypy_cache/
+.ropeproject/
--- a/Currency.py
+++ b/Currency.py
@@ -43,11 +43,12 @@ def handleQuery(query):
             for url in urls:
                 try:
                     response = urlopen(url)
-                except urllib.error.HTTPError as e:
-                    response = e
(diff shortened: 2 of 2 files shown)


g-walsh:
Sorry, I meant to add this to the existing PR. Done now.

ManuelSchneid3r:
Why has this been closed?

g-walsh:
I closed it and made it a PR on #33 (as it was a similar implementation and I didn't want to create two PRs that fixed the same thing) which they never merged into their code. I see you've now merged and changed to another data provider anyway!

g-walsh:
I agree with you that ECB will be much more reliable than scraping from a search provider. As you said it looks like it has fewer available currencies and only updates daily (inferred from the title of the xml).

Might it be more useful to have Yahoo or XE as the first thing to go for and then fall back to ECB as a reliable option? Though it is probably simpler to maintain if it defaults to ECB.

PR #39, opened 2018-05-12
Add an offline emoji picker

SDEagle:
I have implemented a simple offline emoji picker in https://github.com/SDEagle/albert_emoji_picker
Currently it's pretty simplistic, but it does the job. Since there is already another emoji picker in here (requires internet connection though) would there be any interest in having it here? I'm open to moving it into a submodule/continuing developing here or whatever...


ManuelSchneid3r:
Sure. Are there packages containing the alias lists?

SDEagle:
There seems to be https://github.com/carpedm20/emoji/ but appearently their doing pretty much the same thing I did: Scraping the unicode website - though they use http://www.unicode.org/emoji/charts/full-emoji-list.html and I used https://unicode.org/Public/emoji/11.0/emoji-test.txt

ManuelSchneid3r:
The idea of not having to download things every query is fine but maybe getting it initially and cache it would be a more flexible approach.

ManuelSchneid3r:
Arent you interested in making this a bit more sophisticated and flexible? Since I'd really like to have this extension, The online versions are to laggy and not really precise.

SDEagle:
Yes, I certainly am, but right now I have not time at all for this project. I might get to continue a little on it after Christmas. Thing is, despite being so simplistic it actually works well enough at least for me right now... But I will see if I can at least make it fetch the list initially online at the end of the year.

ManuelSchneid3r:
> Thing is, despite being so simplistic it actually works well

Sure but maintenance is a pain. Unicode 12 has been releases last month. Updating would mean to change almost every line of the module. When fetching+parsing is implemented we would have to change a single line only. Even better would be to make it get the current version of unicode somewhere and build the link to fetch the emojilist. In this case we would never ever have to update the script.

SDEagle:
Updated the project to not use a hardcoded list anymore.

It will attempt to fetch the latest emoji list from `https://unicode.org/Public/emoji/latest/emoji-test.txt`. Sadly, that still points to Unicode 11. To get around that and as an offline fallback an initial list (based on Unicode 12) comes along and before updating anything the extension will check if the update is actually newer.

PR #40, opened 2018-05-28
Add jetbrains project finder

mqus:
This plugin searches through the last opened projects of jetbrains products (e.g. CLion, PyCharm, Intellij IDEA, etc) and then displays them when searching for it. When clicking on an option, it opens the project within the appropriate IDE.

This works for the IDEs I have installed on Archlinux and should also work without or with just a few changes on other distributions.


ManuelSchneid3r:
Cool ty :+1:

PR #42, opened 2018-07-03
[pidgin] Add pidgin extension

greizgh:
Let me know if there is anything to rework.

I've tested this extension with a fairly large contact list (~800 contacts) without any problem.


ManuelSchneid3r:
Looks pretty sane to me. Thank you 👍

PR #43, opened 2018-07-04
look for plain 'idea' command on the path

matthewbogner:


mqus:
As the original author of the plugin I have absolutely no issues with this, thank you for your contribution and for using the plugin!

As I have no write-access I can only recommend to merge this.

There is only one thing i'm curious about: Which distribution are you using?

matthewbogner:
@mqus I am using Intellij Ultimate Edition 2018.1 on Ubuntu 18.04 LTS.

PR #44, opened 2018-07-13
timestamp-converter plugin

matthewbogner:
Get the current timestamps:
image

Translate a given timestamp in either millis since epoch or seconds since epoch:
image
image

I do still need a hand figuring out how to include a python module in this plugin that doesn't come with a standard python install.  I have the dateparser module available in the system default python install, but it still doesn't seem to be found when launching albert. 
image


ManuelSchneid3r:
Have you noticed https://github.com/albertlauncher/python/blob/master/DateTime.py?

matthewbogner:
Yes.  I ran with the assumption that it operated differently enough that you may not want it to muddy up the DateTime.py.  I've pushed a change to merge the two. 

Primarily I'm after having a millis since epoch (not just seconds - which is what DateTime.py supported) converted to particular format in both local and UTC time. 

My original question still remains, on how to make use of a python module (dateparser) in this case, so that you can give it any random formatted date and it will return the millis and seconds since epoch.

matthewbogner:
Any additional thoughts? Am I too out on the fringe?  Some other approach you would prefer?

DateTime.py
```diff
@@ -26,7 +26,7 @@ def handleQuery(query):
     if fields:
         def makeItem(text: str, subtext: str):
             return Item(
-                id=__prettyname__,
+                id=text,
```
ManuelSchneid3r: This id not a good idea. The id is used for the ranking and should be somewhat persitent. Using the timestamp effectively messes up the MRU score cache.

matthewbogner:
Good to know.  Fixed.

mqus:
Just to chime in on your problem: I'm running arch and if I install either python-dateparser (for python3) with my package-manager or dateparser with `sudo pip3` it works just fine in both cases.

PR #46, opened 2018-08-06
Add VPN python extension

janeklb:
trigger by starting to type `connect` or `disconnect`, optionally followed by the name of the vpn connection profile

inspired by https://github.com/albertlauncher/python/pull/6

### DIFF
--- a/VPN.py
+++ b/VPN.py
@@ -1,0 +1,66 @@
+"""VPN Extension
+
+Connect or disconnect from a network manager VPN profile"""
+
+
+import subprocess
+from collections import namedtuple
+from albertv0 import *
+from shutil import which
+
+
+__iid__ = "PythonInterface/v0.2"
+__prettyname__ = "VPN"
+__version__ = "1.0"
+__trigger__ = "vpn "
+__author__ = "janeklb"
+__dependencies__ = ['nmcli']
+
+
+iconPath = iconLookup('network-wireless')
+if not iconPath:
+    iconPath = ":python_module"
+
+VPNConnection = namedtuple('VPNConnection', ['name', 'connected'])
+
+
+def getVPNConnections():
+    consStr = subprocess.check_output(
+        'nmcli -t connection show',
+        shell=True,
+        encoding='UTF-8'
+    )
+    for conStr in consStr.splitlines():
+        con = conStr.split(':')
+        if con[2] == 'vpn':
+            yield VPNConnection(name=con[0], connected=con[3] != '')
+
+
+def buildItem(con):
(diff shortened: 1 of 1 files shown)


janeklb:
Screenshot:
image

ManuelSchneid3r:
Actually yes the _trigger_ has to be one word. But the trigger is just to decide if the ext should be executed only on specific prefixes. Its perfectly fine to always run the extension an parse for several prefixes (unless you dont do heavy things besides that check). However I thing (dis-)connect is to broad. I could mean connect wifi, connect ethernet, connect mobile data, connect vpn, connect to facebook :D whatsoever. I thing it may be a good idea to use the trigger *vpn* and use dis-/connect in the title of the item depending on the state of the connection, i.e. when disconnected show "Connect", when connected show "Disconnect". Theres no point in disconnecting a disconnected connection.

janeklb:
Ah yes good point. I’ll use something more specific.

janeklb:
Pushed an update. Trigger is now `vpn `, and automatically detects whether the connection is active or inactive. Downside to this approach is that it's a bit laggy since it retrieves available connections every time it handles the query.

Is it possible to determine, in `handleQuery`, whether we're handling a new albert invocation (ie. first time the trigger was matched while the albert query box was open) or if it's a continuation of an existing one (ie. otherwise)? If so, could cache the connections and invalidate the cache whenever that flag reads "new".

janeklb:
image

janeklb:
I don't think i'll have time to improve upon this PR in the near future.

ManuelSchneid3r:
Just for future reference: 
```
```

janeklb:
Tried this out - it's actually not much faster than using `nmcli`. The main bottleneck is calling `settings_connection.GetSettings()` every time (which is done to determine if the connection is active/connected).

Would performance improve if `handleQuery` returned a generator instead? Or do you read the entire result before doing anything with it?

Otherwise, I would suggest keeping it as is (using `nmcli` to build the list), and then improving if you ever manage to add the suggestion in my [previous comment](https://github.com/albertlauncher/python/pull/46#issuecomment-414126610).

janeklb:
I've also managed to crash `albert` with the dbus approach by spamming it, so it would need some debouncing to ensure stability.

ManuelSchneid3r:
There is a new [community repository for python extensions](https://github.com/albertlauncher/python_community) now. Please send the PR there. Sorry for the inconvenience. I am just trying to reduce the workload here and make you happy with contemporary merges. If you want to join the party (write access) ask for it in the community chats.

PR #48, opened 2018-08-28
Add Maim.py (maim, slop, imgur, notify-send)

balta2ar:
This extension is heavily based on Scrot with the following benefits:
* uses maim and slop -- the latter allows for more consistent and reliable selection (see slop github for the details)
* supports optional uploading to imgur
* supports optional delay that can be specified in the command
* automatically opens browser with the uploaded Imgur URL

### DIFF
--- a/Maim.py
+++ b/Maim.py
@@ -13,11 +13,21 @@ Screenshots will be saved in XDG_PICTURES_DIR or in the temp directory."""
 
 import os
 import subprocess
+import multiprocessing
 import datetime
 import tempfile
+from functools import partial
 from shutil import which
 
 from albertv0 import *
+
+import dbus
+
+
+ACTION_SCREENSHOT_AND_COPY = 0
+ACTION_UPLOAD_TO_IMGUR_AND_COPY = 1
+ACTION_OPEN_IN_BROWSER = 2
+
 
 __iid__ = "PythonInterface/v0.1"
 __prettyname__ = "maim/slop screenshot utility"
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:
Cool 👍. You could use the dbus python package for notifications, I guess dbus is more likely to be present on most systems. [Heres](https://wiki.debianforum.de/Desktop-Notification_von_Systemservice_mittels_dbus#ohne_Erkennung_der_Variablentypen_in_der_Message) an example.

balta2ar:
Thanks! Currently notify-send is used in the middle of command chain. If we want to replace notify-send with a python call to dbus API, I guess it would be necessary to utilize multiprocessing.Process, run all commands there, wait for their completion and send dbus message. Otherwise, how do I put python code in between bash commands? Is that approach OK with you?

balta2ar:
I found this (http://cheesehead-techblog.blogspot.com/2009/02/five-ways-to-make-notification-pop-up.html) and tested method 4, it's working fine on my home machine.

balta2ar:
Refactored a bit and employed dbus module for notifications.

balta2ar:
Weird things started to happen on my home machine. I can't take a screenshot using a selection. It's fine if I just click and select the whole window or if I take a screenshot of the whole screen, but rectangular selection has problems. If albert is popped up on the same monitor (I have two monitors) where I later do rectangular selection, for sufficiently large selections it creates incomplete screenshots with transparent background, e.g.:

2018-08-30t23 43 44 456201-maim-screenshot

It feels like a race condition or something. As if maim allocated the space but couldn't copy the whole picture data...

If, on the other hand, albert pops up on one monitor and I do selection on another one, maim terminates (or slop, that is called by maim, I'm not 100% sure) with the following error:
```X Error of failed request:  BadValue (integer parameter out of range for operation)                                                                                
  Major opcode of failed request:  53 (X_CreatePixmap)                                                                                                             
  Value in failed request:  0x0                                                                                                                                    
  Serial number of failed request:  43                                                                                                                             
  Current serial number in output stream:  47
```

And I only see such behavior when I run albert. If if run parts of this plugin as a regular Python script, everything is fine. Have you seen anything resembling that before?

Maim.py
```diff
@@ -0,0 +1,244 @@
...
+import dbus
+
+
+__iid__ = "PythonInterface/v0.1"
+__prettyname__ = "maim/slop screenshot utility"
+__version__ = "1.0"
+__trigger__ = "maim "
+__author__ = "Yuri Bochkarev"
+# imgur is optional
+__dependencies__ = ["maim", "slop", "notify-send", "xclip"]
```
ManuelSchneid3r: "notify-send"

Maim.py
```diff
@@ -0,0 +1,244 @@
+"""
+Extension which basically wraps the command line utility maim to make
+screenshots from albert. The extension supports taking screenshots of the whole
+screen, an specific area or the current active window.
+
+Also supports upload to imgur:
+https://github.com/tangphillip/Imgur-Uploader/blob/master/imgur
+
+When the screenshot is made, notify-send message will be sent to indicate that
```
ManuelSchneid3r: "notify-send"

balta2ar:
I've just tried it again on Ubuntu 16 machine, and it's working fine... I don't know what's wrong with my Arch machine...

ManuelSchneid3r:
There is a new [community repository for python extensions](https://github.com/albertlauncher/python_community) now. Please send the PR there. Sorry for the inconvenience. I am just trying to reduce the workload here and make you happy with contemporary merges. If you want to join the party (write access) ask for it in the community chats.

PR #45, opened 2018-07-21
Add more binary names

fionera:

PR #49, opened 2018-09-21
[jetbrains] Use jetbrains official icon as fallback

tomsquest:
**Feature**: given no Jetbrains product icons could be found, then use the official icon.

albert-fallback

Also:

- Fix the `dname` variable not being used for the icon (line 40)
- The logo seems to be usable in Albert according to the rules at https://www.jetbrains.com/company/brand/
- Moved `jetbrains-projects.py` to its own folder, with the same name: the naming is now uniform as other directories are CamelCased, this can be changed but users will have to re-enable this plugin


tomsquest:
@mqus I would like to make this modifications before merging:
- loop only once on `namestr`
- attempt to use a `for else`
- pep8 format

tomsquest:
@mqus Here how `find_exec` could be rewritten:

```python
default_icon = os.path.dirname(__file__) + "/jetbrains.svg"

def find_exec(namestr: str):
    for name in namestr.split(" "):
        exec = which(name)
        if exec:
            icon = iconLookup(name) or default_icon
            return exec, icon
    else:
        return None
```

Question: this code only tries to find an icon named after the binary (eg. given 'pycharm-eap' is found, then an icon named 'pycharm-eap' is searched for), so could it be possible that a binary won't have the corresponding icon name?
If this is the case, the code has to loop twice, one for searching the binary, and when found, loop on all possible icon names, returning the fallback when missing.


BTW, my setup is this one:
- Intellij extracted in `~/Apps/idea/current`
- `$PATH` included the `idea` launcher script (the one created by Intellij and which allows to open a file, this script refers to `~/Apps/idea/current`)
So I don't have any jetbrains icons in my path (hence this PR for having a fallback icon)

mqus:
Thanks! I think the comments are far better than before now :D 

One last minor nitpick I noticed just now (you can choose if you want to change that): the for-else-pattern was originally meant for `break`s inside the loop, not `return`s. The way it's written right now makes the `else` uneccessary, as the `return None` is called everytime the for loop finishes anyway. So my suggestion would be to remove the `else:` line and just return None at the end of the function normally. But that is just a formal issue, this can be merged as-is.

tomsquest:
Thanks @mqus, I removed the unnecessary else.

BTW I found two issues that I will try to fix in other PRs:
- Sometimes, just after launching Albert, and typing the first letter of `jb`, this error is displayed `TypeError: 'NoneType' object is not subscriptable` in __init__.py(102), and the extension does not work
- Given a fresh install of Pycharm EAP: `FileNotFoundError: [Errno 2] No such file or directory: '/home/tom/.PyCharm2018.3/config/options/recentProjectDirectories.xml` in __init__.py(59)

I am also concerned about the code finding the iconName depending of the binary. I think that the icon will be always the same name (`idea.png` even if the executable is name `idea-ce-eap` or anything not `idea`). I am trying to find an elegant way to code this second loop on the names.

tomsquest:
@mqus Should I squash the commits?

tomsquest:
@mqus I fixed an issue when the extension was fired even when `jb` was not entered.
I also fixed the case of a missing recentProjects file.

mqus:
Regarding the squashing: That's for @ManuelSchneid3r to decide, I don't have merge rights ;-)

mqus:
Regarding your issue, I think we should only show projects where find_exec doesn't return None, meaning we should insert a 
```python
            if not binaries[app[0]]:
                continue
```
At line 92. 

This will also mean that we ignore all jetbrains executables which are not in the path variable and I think this is fine, as we can't reasonably search the whole filesystem for the binaries. If something is not in the PATH variable, then put it inside it or create symlinks inside of a directory that is included in PATH. 

We could search for the binaries *once* and then cache the locations, but I'm against this because it's a hacky workaround. 

In this case a plugin settings UI might be the most reasonable way to do this, but sadly we don't (yet) have that option for python plugins.

PS: maybe the path to the executable is contained in the config files somewhere? I'll do some research...

tomsquest:
Ok. I was rewriting the `Items` for-comprehension to pinpoint the problem.
I was thinking the same: no need to read projects when the tool has no binary in path.

I have two questions:

1. from where the binary names like `pycharm pycharm-eap charm` come from? In a fresh Pycharm EAP install, I only found `pycharm.sh` and it can create a launcher named `charm`. So do you think we should add `pycharm.sh` and can `*-eap` be removed?
1. more general, can you point me to a doc on how to import `albertv0` from Python code? I wrote some tests and they need to import this.

Thanks @mqus

mqus:
1. The binary names come from packages users can install. As I'm an archlinux user I simply searched through the [AUR](https://aur.archlinux.org/packages/?O=0&SeB=nd&K=pycharm&outdated=&SB=n&SO=a&PP=50&do_Search=Go) and looked at the PKGBUILDs used (the makefile for archlinux packages). E.g. the [PKGBUILD of pycharm-eap](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=pycharm-eap#n79) creates a symlink to pycharm.sh and calls it pycharm-eap. There are also #43 and #45 which added more names. You can add the default names (`charm` and `pycharm.sh`) if they're missing for you. I simply gathered binary names that were used.

2. If you mean that you want to run python scripts importing `albertv0` outside of albert then I'm also clueless. But I *think* that `albertv0` is a (pseudo-)module which is provided by albert at runtime and you can't use it externally.

tomsquest:
Just refactored to give some meaning to the variables.

jetbrains-projects/__init__.py
```diff
@@ -93,14 +93,33 @@ def handleQuery(query):
...
+            project_path = p[1]
+            project_dir = project_path.split("/")[-1]
+            product_name = p[2]
+            binary = binaries[product_name]
+            if not binary:
+                continue
+
+            executable = binaries[p[2]][0]
+            icon = binaries[p[2]][1]
+
```
mqus: I think we should reuse `binary` here (e.g. `executable = binary[0]`)

tomsquest: :heavy_check_mark: done

tomsquest:
@mqus before merging, I think I should capitalize the name of the directory, ie. from `jetbrains-projects` (named after the original Python file) to `JetbrainsProjects`.
All other extensions are Capitalized, same for the directories.

Confirm?

tomsquest:
:heavy_check_mark: rename plugin dir to `JetbrainsProjects`

PR #50, opened 2018-10-18
Add Fortune extension

netsgnut:
This PR proposes adding a `fortune` extension. When installed and enabled, `fortune` will `print a random, hopefully interesting, adage`. Selecting it copies the line to clipboard.

### DIFF
--- a/Fortune.py
+++ b/Fortune.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+
+"""Fortune wrapper extension"""
+
+from albertv0 import *
+from shutil import which
+import subprocess as sp
+
+__iid__ = "PythonInterface/v0.1"
+__prettyname__ = "Fortune"
+__version__ = "1.0"
+__trigger__ = "fortune"
+__author__ = "Kelvin Wong"
+__bin__ = "fortune"
+__dependencies__ = [__bin__]
+
+if which(__bin__) is None:
+    raise Exception("'%s' is not in $PATH." % __bin__)
+
+iconPath = iconLookup("font")
+
+
+def handleQuery(query):
+    if query.isTriggered:
+        newFortune = generateFortune()
+        if newFortune is not None:
+            return getFortuneItem(query, newFortune)
+
+
+def generateFortune():
+    try:
+        return sp.check_output(["fortune", "-s"]).decode().strip()
+    except sp.CalledProcessError as e:
+        return None
+
+
+def getFortuneItem(query, fortune):
+    return Item(
+        id=__prettyname__,
+        icon=iconPath,
+        text=fortune,
+        subtext="Copy this random, hopefully interesting, adage",
+        completion=query.rawString,
+        actions=[ClipAction("Copy to clipboard", fortune)]
+    )


Fortune.py
```diff
@@ -0,0 +1,41 @@
...
+    item = Item(
+        id=__prettyname__,
+        icon=iconPath,
+        text=__prettyname__,
+        completion=query.rawString
+    )
+    try:
+        item.text = sp.check_output(['fortune', '-s'],
+                                    stderr=sp.STDOUT).decode().strip()
+        item.subtext = "Copy this random, hopefully interesting, adage"
```
tomsquest: `subtext` could be set directly in `Item()`

Fortune.py
```diff
@@ -0,0 +1,41 @@
...
+def generateFortune(query):
+    item = Item(
+        id=__prettyname__,
+        icon=iconPath,
+        text=__prettyname__,
+        completion=query.rawString
+    )
+    try:
+        item.text = sp.check_output(['fortune', '-s'],
+                                    stderr=sp.STDOUT).decode().strip()
```
tomsquest: Why `stderr=sp.STDOUT`?

netsgnut: Indeed it should be not required. Was fiddling around the error handling but left this line there.

Fortune.py
```diff
@@ -0,0 +1,45 @@
...
+    if query.isTriggered:
+        newFortune = generateFortune()
+        if newFortune is not None:
+            return getFortuneItem(query, newFortune)
+
+
+def generateFortune():
+    try:
+        fortuneOutput = sp.check_output(["fortune", "-s"]).decode().strip()
+        return fortuneOutput
```
tomsquest: `fortuneOutput` can be inlined

Fortune.py
```diff
@@ -0,0 +1,45 @@
...
+
+__iid__ = "PythonInterface/v0.1"
+__prettyname__ = "Fortune"
+__version__ = "1.0"
+__trigger__ = "fortune"
+__author__ = "Kelvin Wong"
+__dependencies__ = ["fortune"]
+
+if which("fortune") is None:
+    raise("'fortune' is not in $PATH.")
```
tomsquest: Tried the extension **without** fortune being installed, got this error:

```bash
12:10:56 [WARN] [Fortune.py] TypeError: exceptions must derive from BaseException

At:
  ./Fortune.py(17): <module>
  <frozen importlib._bootstrap>(219): _call_with_frames_removed
  <frozen importlib._bootstrap_external>(678): exec_module
```

So I think `raise` is not enough here for proper error handling.

Fortune.py
netsgnut: True. Updated with a9f52a6. Thanks.

ManuelSchneid3r:
Cool. :+1:

ManuelSchneid3r:
I am somehow thinking about writing a generic CLI wrapper that allows to specify a trigger and a commandline. A lot of python extensions are basically CLI wrappers. So this would make things easier I guess.

netsgnut:
Would love to see that happen. With that it sounds many of the boilerplate code can be deprecated.

PR #51, opened 2018-10-22
Coinmarketcap plugin: better price formatting

sharipov-ru:
PR changes price formatting for coinmarketcap plugin by using precision based on prices. 

As we can see on [http://coinmarketcap.com](http://coinmarketcap.com), they never show BTC value as 6485,481567 - such precision is not needed for coins with price greater than $1.

Before:
screenshot_20181022_214806

After:

screenshot_20181022_214721

### DIFF
--- a/CoinMarketCap/__init__.py
+++ b/CoinMarketCap/__init__.py
@@ -73,7 +73,8 @@ class UpdateThread(Thread):
                     vol = coindata['24h_volume_usd']
                     vol = lformat("%d", float(vol), True) if vol else "?"
                     price = coindata['price_usd']
-                    price = lformat("%f", float(price), True) if price else "?"
+                    price_precision = "%.2f" if float(price) > 1 else "%.6f"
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:
I have to say that I am not really satisfied with this solution since the significant figures vary to much. take eos in the example with 3 s.f. and tether with 6 s.f.

ManuelSchneid3r:
Maybe this is a good solution. https://stackoverflow.com/a/3411435/978486. We still need to have a consens on a useful amount of significant figures.

sharipov-ru:
@ManuelSchneid3r 

since it's a coinmarketcap extension, I thought it would be better just to show all values in the same format as they currently listed on coinmarketcap (and it seems they use this "more or less $1 rule" to set the precision). So that with the current logic in PR, all values are identical to coinmarketcap values.

ManuelSchneid3r:
Well it is better than before, so what 👍 ty

PR #53, opened 2018-11-13
Copyq improvements

Tyilo:
For the last commit ("Properly handle escaping of underlined texts"), the problem before can be seen by copying `<foo> "bar"` and then searching for the empty string.

### DIFF
--- a/CopyQ.py
+++ b/CopyQ.py
@@ -12,7 +12,7 @@ from albertv0 import *
 
 __iid__ = "PythonInterface/v0.1"
 __prettyname__ = "CopyQ"
-__version__ = "1.0"
+__version__ = "1.1"
 __trigger__ = "cq "
 __author__ = "Manuel Schneider"
 __dependencies__ = ["copyq"]
--- a/CopyQ.py
+++ b/CopyQ.py
@@ -57,7 +57,6 @@ def handleQuery(query):
     if query.isTriggered:
         try:
             pattern = re.compile(query.string, re.IGNORECASE)
-            critical(query.string)
         except re.error:
             return [
                 Item(
--- a/CopyQ.py
+++ b/CopyQ.py
@@ -93,7 +93,11 @@ def handleQuery(query):
             if not text:
                 text = "<i>No text</i>"
             else:
-                text = pattern.sub(lambda m: "<u>%s</u>" % m.group(0), html.escape(" ".join(filter(None, text.replace("\n", " ").split(" ")))))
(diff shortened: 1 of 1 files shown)


CopyQ.py
```diff
@@ -69,7 +69,8 @@ def handleQuery(query):
                 )
             ]
 
-        script = copyq_script_getMatches % query.string if query.string else copyq_script_getAll
+        escaped_query = query.string.replace('\\', '\\\\').replace('"', '\\"')
```
ManuelSchneid3r: why is this necessary?

CopyQ.py
```diff
@@ -55,21 +55,48 @@
...
             text = json_obj['text']
             if not text:
                 text = "<i>No text</i>"
             else:
-                text = pattern.sub(lambda m: "<u>%s</u>" % m.group(0), html.escape(" ".join(filter(None, text.replace("\n", " ").split(" ")))))
+                U_BEGIN = "@U_BEGIN_XXXXX@"
+                U_END = "@U_END_XXXXX@"
+                raw_text = " ".join(filter(None, text.replace("\n", " ").split(" ")))
+                underlined_text = pattern.sub(lambda m: U_BEGIN + m.group(0) + U_END, raw_text)
+                text = html.escape(underlined_text).replace(U_BEGIN, "<u>").replace(U_END, "</u>")
```
ManuelSchneid3r: why this indirection over U_BEGIN and U_END?

CopyQ.py
Tyilo: Otherwise the `script` will not have valid js syntax if the query includes `"` or `\\`.

CopyQ.py
Tyilo: The problem is that if you have copied `<u></u>` (and other tags), we want to escape them so we don't view them raw.
We still want to add `<u></u>` to the parts of the string that matched the query, so we need this indirection.

ManuelSchneid3r:
Damn I just went through it to notice that I screwed up by moving them to dirs. If you could fix that I'll merge instantly

ManuelSchneid3r:
tely I let you all wait pretty long. Now I have the problem that I don't know which PR's are valid at all. Therefore I will close all PR's for tentatively. If you want me to reopen the PR let me know. Also it would be nice if you could resolve potential conflicts due to changes in the repo. Sorry for the inconveniences, but I am working on it for a week now. I'm in a hurry. Note that this is a copy paste message. 😉

PR #55, opened 2018-12-22
Added new Currency provider

imrancio:
Yahoo provider for Currency converter does not work anymore, thus only returning few results from European Central Bank. I replaced it with a free API call to currencyconverterapi.com. Also added some usage instructions for the extension.

### DIFF
--- a/Currency.py
+++ b/Currency.py
@@ -3,6 +3,7 @@
 """Convert currencies.
 Synopsis: <amount> <src currency> [to|as|in] <dest currency>"""
 
+import json
 import re
 import time
 from urllib.request import urlopen
--- a/Currency.py
+++ b/Currency.py
@@ -1,10 +1,8 @@
 # -*- coding: utf-8 -*-
 
 """Convert currencies.
-
 Synopsis: <amount> <src currency> [to|as|in] <dest currency>"""
 
-import json
 import re
 import time
 from urllib.request import urlopen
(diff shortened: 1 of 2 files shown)


ManuelSchneid3r:
bildschirmfoto vom 2018-12-23 18-06-52

works well here

Currency.py
```diff
@@ -50,34 +52,41 @@ def convert(self, amount, src, dst):
             dst_rate = self.exchange_rates[dst]
             return str(amount / src_rate * dst_rate)
 
-class Yahoo:
+class Api:
```
ManuelSchneid3r: please give this a reasonable name an reintroduce yahoo.

Currency.py
```diff
@@ -14,6 +15,7 @@
 __iid__ = "PythonInterface/v0.1"
 __prettyname__ = "Currency converter"
 __version__ = "1.0"
+__trigger__ = "cc "
```
ManuelSchneid3r: why do you introduce a trigger? actually it was intended to work without the trigger.

Currency.py
imrancio: Can do. Yahoo does not work for me. I tested the regex on the html that was returned to me and it was redirected to Yahoo AU. So there may be inconsistencies in results across regions and the Yahoo Finance API has been deprecated for a while.
screenshot_2018-12-24_16-48-30

imrancio:
screenshot_2018-12-24_17-34-34
It works, but only with proper currency codes. Yahoo would have been nice for more general search.

imrancio:
> bildschirmfoto vom 2018-12-23 18-06-52
> 
> works well here

Also I don't know if we're looking at the same currency, but your result seems way off.

imrancio:
I guess this will never be merged in. At least, the other 2 currency providers have been fixed now.

PR #56, opened 2019-01-02
Add mpd control extension

mqus:
This adds a small extension to control mpd (music player deamon) via albert. Currently one can search and add songs, albums and all songs of an author. 
To change the controlled instance, one has to edit the SERVER and PORT variables.

It needs the python package [mpd2](https://github.com/Mic92/python-mpd2)


mpd/__init__.py
```diff
@@ -0,0 +1,215 @@
+import os
```
ManuelSchneid3r: plase add a docstring containing a description and a synopsis. synopsis is important since else nobody can use the extension.

mpd/__init__.py
```diff
@@ -0,0 +1,215 @@
...
+__prettyname__ = "mpd: control your MPD"
+__version__ = "1.0"
+__trigger__ = "mpd "
+__author__ = "Markus Richter"
+__dependencies__ = []
+
+SERVER = "localhost"
+PORT = 6600
+
+# All included icons are part of the papirus icon theme by the Papirus Development Team and are licensed under the GPLv3.
```
ManuelSchneid3r: please use the packed icons only as fallback use albertv0.iconlookup() to find themed icons.

mpd/__init__.py
```diff
@@ -0,0 +1,215 @@
+import os
+from albertv0 import *
+from mpd import MPDClient
+
+__iid__ = "PythonInterface/v0.1"
+__prettyname__ = "mpd: control your MPD"
```
ManuelSchneid3r: This is the pretty name. I guess "Music Player Daemon" would make sense. Put the brief description into the docstring.

mpd/__init__.py
mqus: The problem here is that some of these icons are somewhat misused. Eg. the icon representing an artist collection is an icon which usually is named "folder-home" or "user-home". As far as I'm aware, there is no canonical icon for artists. If would choose to select the "folder-home" icon of the current system icon set, there would probably pop up a folder with a house on it which would be a really bad choice for an icon representing an artist. A similar, but smaller case can be made for the "folder-music" icon representing an album... 

As for the single songs, I'll choose "audio-x-generic" which seems to be a good fit.

mpd/__init__.py
mqus: Sure!

mpd/__init__.py
mqus: As an aside: where(and how) does iconLookup look for Icons?

mqus:
As an aside: where(and how) does iconLookup look for Icons?

ManuelSchneid3r:
Google freedesktop icon lookup

mqus:
I chose "audio-x-generic", "music-album-default" and "music-artist-default" for the icon names... If this is ok, this would be ready to merge from my side.

ManuelSchneid3r:
There is a new [community repository for python extensions](https://github.com/albertlauncher/python_community) now. Please send the PR there. Sorry for the inconvenience. If you want to join the party (write access) ask for it in the community chats.

PR #58, opened 2019-01-27
Added Snip - a SCReenshOT alternative

thepante:
Why another? In this case uses gnome-screenshot utility. 

With `scrot` I'm getting artifacts when use of the area function (globally, not albert related). Quite often I do screenshots, so I did this one that is focused on _taking snips of the screen_ but can take normal screenshot and so on... So with this I get what I was looking for.

Here is the [readme](https://github.com/thepante/albert-snip/blob/master/README.md) on my repo for the extension


ManuelSchneid3r:
There is a new [community repository for python extensions](https://github.com/albertlauncher/python_community) now. Please send the PR there. If you want to join the party (write access) ask for it in the community chats.

PR #60, opened 2019-02-24
Fix installation command

shenxn:
The tilde `~` in double-quotes are not going to be expanded ([Double-Quotes](http://pubs.opengroup.org/onlinepubs/009604599/utilities/xcu_chap02.html#tag_02_02_03)). It causes the repository to be cloned inside a subdirectory named `~` instead of the home directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ This repository is shipped with albert. If you want to have bleeding edge extens
 To install the extensions in user space type the following in your terminal:
 
 ```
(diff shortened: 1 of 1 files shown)


tomsquest:
Verified on my side (bash/zsh).

@shenxn What about using `$HOME`, it is correctly expanded with double quotes, and double quotes protect against spaces in the string.

shenxn:
@tomsquest That works as well, but I don't think it's necessary since there is no space in the target directory.

PR #62, opened 2019-03-05
Fixed icon of Dango Emoji extension

BobbyWibowo:
Explicitly specify the file name, just like how it's done in Dango Kao extension.
Previously the resulting file name would `albert.dango_emoji.png`, which was obviously wrong.


BobbyWibowo:
Was there some sort of forgetting to push for months thingy happening or something? lmao
Cause like I just noticed that for some reason my base was outdated, but I was sure it wasn't when I made this PR

PR #63, opened 2019-04-15
Google Suggest extension

HansCz:
Let me know what you think :)


google_suggest/__init__.py
```diff
@@ -0,0 +1,88 @@
...
+user_agent = "Mozilla/5.0"
+limit = 20
+
+
+def initialize():
+    global baseurl
+
+def handleQuery(query):
+    if query.isTriggered:
+
```
djmbritt: I would consider adding:
sleep(0.2)
if not query.isValid:
            return

That way avoid being rate-limited by google.
I've tried it with a couple of other google end points, and after a while they always end up limit the number of calls you can make to their API endpoint.

djmbritt:
Except for the comment I made about constantly making calls to the URL, I really like this. Works great.

google_suggest/__init__.py
HansCz: I wasn't aware that they rate-limited. Thanks for the heads up.
I'll put it in.
Will also go see if I can find specifics on how the rate limiting functions, so I can add a warning in a comment for anybody who wants to alter the code.

HansCz: I added your suggested change. 

Unfortunately I was unable to find any definite numbers on when and how the rang-limiting kicks in, so I left it out.

HansCz:
Have you considered whether to merge this?

djmbritt:
I'd like for this to be merged. But I don't have write access. 
I do know that the Manuel is busy at the moment and doesn't have time. 

The python-community repo is open though, for other community members to merge pull requests.

ManuelSchneid3r:

PR #65, opened 2019-04-18
Add Network Manager VPN plugin

igorolhovskiy:
Using nmcli and subprocess
Able to turn on/off VPN from Albert interface with `vpn` trigger


ManuelSchneid3r:
Merged and moved from comm repo

PR #69, opened 2019-08-26
Add 'notify-send'.

tehe:
Hi.

I thought that adding notify-send to the timer module in the Python plugin will make things a little better than just playing the sound. I've tested only on Linux Fedora 30, in Gnome.

Thanks,
Tomas


ManuelSchneid3r:

besides: this adds a dependency. I think this may also be achievable with dbus

PR #72, opened 2019-11-15
fix: currency converter round float returns value times 10

grafuls:
When querying for a rounded amount like "400 usd to ars", the value
passed for the yahoo query comes through as 400.0, which is in turn
interpreted by the yahoo search api as 4000.
Added formating and conversion on the api query string generation
for passing 2 decimal places.


ManuelSchneid3r:

PR #74, opened 2019-11-21
Define path and binary for RubyMine

dsager:
RubyMine is another IDE from JetBrains that has the same structure as other IntelliJ based products. So just adding the path for it will make ruby projects appear in the launcher.

For example:
image


mqus:
LGTM, but sadly I have no merge rights and the project isn't very active... (I wrote the initial version of the plugin and seem to have missed RubyMine...)

mqus:
Short notice for everyone who is interested: I made a few more fixes to the plugin and am now hosting the plugin here: https://github.com/mqus/jetbrains-albert-plugin

dsager:
Great, I installed it and it works like a charm. Thanks! :+1:

PR #80, opened 2020-02-13
Linguee

davekch:
This pull request adds a plugin for linguee support as requested in [this issue](https://github.com/albertlauncher/albert/issues/807).
The contents displayed are the same that linguee presents when typing into its search bar.
Screenshots german-english:
Screenshot from 2020-02-13 23-15-26
Screenshot from 2020-02-13 23-13-20
Screenshots english-german:
Screenshot from 2020-02-13 23-15-37
Screenshot from 2020-02-13 23-14-06

The language should be easily changeable by adjusting  `lang`  in the script (not tested).

Selecting an item in albert and hitting enter opens up the corresponding site on linguee.com


ManuelSchneid3r:
Cant you find an svg icon for the ext?

Quick search came up with this

https://d1wigddrwdtsce.cloudfront.net/img5/favicon.svg with 1.817 Byte other wise looks sane. just some reviews needed.

ManuelSchneid3r:

davekch:
Hi, could you please reopen this PR? I think it's a useful extension that could benefit other users as well.

It does not change any existing files, so there are no conflicts.

Thanks!

linguee/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+Synopsis: <trigger> <word>"""
+
+
+from albertv0 import *
+import requests
+import bs4
+import os
+
+
+lang = "deutsch-englisch"
```
ManuelSchneid3r, line 16: is there a way to let the ext recognize this on its own? if upstream then it should be internationlaized.

linguee/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+
+__iid__ = "PythonInterface/v0.1"
+__prettyname__ = "Linguee-deutsch-englisch"
+__version__ = "0.1"
+__trigger__ = "lin "
+__author__ = "Lucky Lukert, David Koch"
+__dependencies__ = ["beautifulsoup4"]
```
ManuelSchneid3r, line 23: i know this comes in handy, but it is annoying for end users and can be avoided (unless I missed something)

linguee/__init__.py
davekch: It would be possible to make the user specify a language like "lin de some word" or "lin it qualche parola". I would however like to keep english as the default second language because linguee might not support translations from any arbitrary language into another and also the prompt would become more cluttered (imagine "lin de it durcheinander").
Would this be okay?

linguee/__init__.py
davekch: ok I'll try to remove this dependency

linguee/__init__.py
```diff
@@ -74,29 +74,28 @@ def get_suggestions(query):
             result["translations"].append(transl_str)
         results.append(result)
 
-    print(results)
+    # print(results)
```
graue70, line 77: There probably shouldn't be commented code in the final version.

ManuelSchneid3r: Why. Rather completely removed

davekch:
ok so I addressed your comments.
I removed the dependency on `beautifulsoup4`.
Also, the extension works now for the languages de (German), es (Spanish), fr (French), it (Italian), ja (Japanese), nl (Dutch), pl (Polish), pt (Portuguese),ru (Russian) and zh (Chinese). These are the languages that linguee can translate to and from english.
The new behavior looks like this:
Screenshot from 2020-11-10 19-58-32
Screenshot from 2020-11-10 19-59-08
Screenshot from 2020-11-10 19-59-26

graue70:
I know you already changed it, but to be honest, I would never use it like it is now. I think that the average use case is a non-English speaking person wanting to translate from English and to English from that one other language. They never need to translate a third language.

That means that the language besides English would have to be configurable somehow. Then the translation should be able to be started with just `lin ...` like you originally proposed.

@ManuelSchneid3r Is there a way to configure a python extension in the settings like a native extension? Or is the only way to change a hard-coded string in the code file?

ManuelSchneid3r:
Unfortunately not. But why not making it a native extension with nice checkboxes and stuff?

davekch:
@graue70 It's still useful for the average user you're describing. Having to type "lin it " is a minor inconvenience compared to just "lin ". I agree that it would be more comfortable to just type "lin " and have the language set via some kind of setting but I would not say that this makes the extension useless / unusable.

If you disagree to the point that you don't see a way to merge this but rather make a native extension, I'd also be willing to work on this but I'd need some help to get started...

graue70:
> @graue70 It's still useful for the average user you're describing. Having to type for example "lin it " is a minor inconvenience compared to just "lin ". I agree that it would be more comfortable to just type "lin " and have the language set via some kind of setting but I would not say that this makes the extension useless / unusable.
> 
> If you disagree to the point that you don't see a way to merge this but rather make a native extension, I'd also be willing to work on this but I'd need some help to get started...

I don't make any decisions as to what is merged and what isn't, so feel free to ignore my input. I'm only stating my honest opinion.

As to the extension. It would obviously be better suited to be a native extension in order to be configurable in the way described. One option to go forward is to merge the python extension and remove it when/if a native extension is finished.

As to python extensions in general, I think that they would profit from a way to configure them somehow. But this isn't the right place to discuss this.

ManuelSchneid3r:
> As to python extensions in general, I think that they would profit from a way to configure them somehow. But this isn't the right place to discuss this.

The app was never intended to fully support python. Sure Python is a nice way to protype and lean to maintain, however Python extensions are slow af and cannot be parallelized/threaded (GIL) or run concurrently (asyncio), unless we find a way to "merge" the asyncio event loop in the Qt event loop. I hoped that finally good Python exts will find their way into a native implementation.

I will release 0.16.2 soon. We can merge this if you want, but also I want tho make C++ exts way smaller and easier to start with. Just let me know.

graue70:
> The app was never intended to fully support python. Sure Python is a nice way to protype and lean to maintain, however Python extensions are slow af and cannot be parallelized/threaded (GIL) or run concurrently (asyncio), unless we find a way to "merge" the asyncio event loop in the Qt event loop. I hoped that finally good Python exts will find their way into a native implementation.

I see. It would probably be a good idea to put this statement into the docs.

davekch:
> I will release 0.16.2 soon. We can merge this if you want, but also I want tho make C++ exts way smaller and easier to start with. Just let me know.

that's great news! Then I'd rather wait for the release and start working on a native extension once it's easier to do so. :)
I use this extension almost everyday and I underestimated how annoying it is to type in the language on every query. So being able to configure it would be really great.

ManuelSchneid3r:
Sorry that I did not hold my promise. I just released 0.16.2 but no changes to the core. My main machine is a mac meanwhile. I have to get the mac port done first to make dev easier. My linux box is in fact still a desktop but not really effective since the hardware just lungers in the coner of my living room and basically just serves as a headless server. On my todolist though.

LuckyLukert:
Wow, it's amazing that my little script is getting that much of attention:) Thanks, @davekch !

I agree with @graue70 that typing the language each time is bad.

One solution I see is to take the system language as default (in combination with English).
And for some other languages, one just can write linfr / linit / linru / ...
Then you can use the normal "lin" for system language to English.

PR #83, opened 2020-09-05
Fix key errors in youtube plugin

dshoreman:
Resolves an issue reported by Laurent Meyer in the telegram chat, caused by a change in the data structure used on Youtube.

Previous data:
```json
{ "simpleText": "3 videos" }
```

Current data:
```json
{ "runs": [
  { "text": "3" },
  { "text": " videos" }
]}
```

It would seem that values are gradually being switched as many still use the `simpleText` key, so to help future-proof things slightly, all values are updated to use a helper function that autodetects either `runs` or `simpleText`.

### DIFF
--- a/youtube.py
+++ b/youtube.py
@@ -20,15 +20,14 @@ __trigger__ = 'yt '
 __author__ = 'Manuel Schneider'
 __icon__ = iconLookup('youtube')  # path.dirname(__file__) + '/icons/YouTube.png'
 
+DATA_REGEX = re.compile(r'^\s*(var\s|window\[")ytInitialData("\])?\s*=\s*(.*)\s*;\s*$', re.MULTILINE)
+
 HEADERS = {
     'User-Agent': (
         'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko)'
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:
@dshoreman is this still valid?

dshoreman:
In its present state, no - there've been further changes to YouTube that break it.

The data relied on by the original implementation is now retrieved through ajax:

Screenshot of Firefox devtools showing JSON returned from youtube.com/search when given a valid key

While `ytInitialData` is no longer available, searching the source for the key in dev tools *does* find a match somewhere within a huge single-line `<script>` block.

I'm in the process of prettifying the js atm so that I can locate the right key. My theory is we should be able to use it to manually send the second `POST` that actually fetches the `ytInitialData` object. Will update when I find something

dshoreman:
@ManuelSchneid3r This is good for review now. Due to an indent change in the original `if match:` block, the diff's a bit challenging so you'll want to view it [with whitespace changes hidden](https://github.com/albertlauncher/python/pull/83/files?diff=split&w=1).

ManuelSchneid3r:
thank you. did not trace down every single path, but it does not look harmful and for some reason you seem trustworthy to me 🤣

PR #89, opened 2020-11-03
feat: add php evaluator plugin

devhammed:
Screenshot from 2020-11-03 17-47-55

This pull request adds the ability to evaluate PHP expressions.


php_eval/__init__.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+
+"""Evaluate simple JavaScript expressions. Use it with care every keystroke triggers an evaluation."""
```
graue70, line 3: That should be php instead of JavaScript.

devhammed: Fixed.

ManuelSchneid3r:
Here as well, please remove fstrings and check if the [binary exists](https://github.com/albertlauncher/python/blob/dev/units/__init_.py#L25-L26).

devhammed:
> Here as well, please remove fstrings and check if the [binary exists](https://github.com/albertlauncher/python/blob/dev/units/__init_.py?rgh-link-date=2020-11-08T23%3A04%3A28Z#L25-L26).

Done.

ManuelSchneid3r:
Cool, thank you. If you like the idea you could also make this directory a dedicated repo (like `albert-php-eval`) in your GH account and send a PR replacing this directory with a git submodule, which is the said repo. Same for the node thing

devhammed:
> Cool, thank you. If you like the idea you could also make this directory a dedicated repo (like `albert-php-eval`) in your GH account and send a PR replacing this directory with a git submodule, which is the said repo. Same for the node thing
> 
> like this [dshoreman/albert-translate](https://github.com/dshoreman/albert-translate)

Oh but I think I am cool with this method because I am still keeping the repo around, the contributions still count anyways.

ManuelSchneid3r:
What do you mean by counts anyway?

devhammed:
> What do you mean by counts anyway?

I mean the contribution to the project will still show in my profile, something like this:

Screenshot from 2020-11-09 16-19-52

PR #91, opened 2020-11-10
improve gnu units output parsing

iyzana:
`units 500seconds time` is currently giving output like this `0;0;0;8;20` which is not very human friendly
this PR changes the output to `8 min + 20 sec`

i believe the other outputs are unchanged:
`units 5min` => `300 s`
`units 1foot cm` => `30.48`

also, giving negative inputs is now supported, as units interpreted them as flags before

### DIFF
--- a/units/__init_.py
+++ b/units/__init_.py
@@ -30,7 +30,7 @@ if not icon:
     icon = ":python_module"
 
 regex = re.compile(r"(\S+)(?:\s+to)\s+(\S+)")
-unitListOutput = re.compile(r"(\d+;)+[\d.]+")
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:
> `units 500seconds time` is currently giving output like this `0;0;0;8;20`

```
~ gunits 500seconds time
	8 min + 20 sec
```

iyzana:
the plugin runs units with the `--terse` flag, in order to get parsable output for most other queries (e.g. `units 500seconds`).
but in terse mode the unit list output is semicolon-seperated.

PR #92, opened 2020-11-11
Adding rand.py script for randomly drawing integers

cyprienruffino:
This is a (very) simple module that allows to sample random numbers from Albert. It allows drawing several integers in a defined range. I use it for dice rolls.


ManuelSchneid3r:
Looks sane

cyprienruffino:
I just removed the icon and updated the code consequently.

ManuelSchneid3r:
well it would be cool though if it had any icon. but iirc if the icon is omitted the unknown icon is used which renders to a question mark, right? this would acutally somehow match the semantics of random.

tomsquest:
@cyprienruffino in order to find an icon, I used in the past iconfinder.ocm with a filter free and a compliant license. 

For example: https://www.iconfinder.com/search/?q=random&price=free&license=gte__3

cyprienruffino:
Added a free icon to the project, moved everything to a dedicated folder. Thanks @tomsquest for the useful link ! :)

PR #94, opened 2020-11-18
Add Stadia plugin

sebastiansam55:
Usage: `stadia [play|store] <game name>`

`play` launches the player webpage for the game selected

`store` launches the store webpage for the game selected

Includes both generated icons for each game and methods used to generate game list and icons.


ManuelSchneid3r:
Ain there a way to fetch the icons and games dynamically?

sebastiansam55:
There is no api for stadia that I am aware of. The regex would also
presumably break anyway if they change the design of the website, and the
regex is also based on an excerpt of the page not the entire page so not
sure if those tags exist elsewhere on the all game page.

I provided the code to extract and fetch/resize/convert images from the
webpage but not completely automated, not even sure how often games are
added to stadia.

ManuelSchneid3r:
The problem I have with this plugin is that it needs massive maintenance. Please make it a git submodule, maintain it and let me know if you do not anymore. This way you have your own issue tracker where ppl can notice you about missing games.

sebastiansam55:
Thats fair. I looked into setting it up so the plugin will refresh automatically but from a brief glance the list of all games I used for the manual generation can only be accessed by someone who is logged into a stadia account (so you wouldn't be able to easily download the webpage from python) I'll swap it to a submodule.

PR #97, opened 2020-12-05
Improve units output

iyzana:
`units 500seconds time` in albert is currently giving output like this `0;0;0;8;20` which is not very human friendly
this PR changes the output to `8 min + 20 sec`

i believe the other outputs are unchanged:
`units 5min` => `300 s`
`units 1foot cm` => `30.48`

also, giving negative inputs is now supported, as units interpreted them as flags before

this time against the master branch

### DIFF
--- a/units/__init__.py
+++ b/units/__init__.py
@@ -30,7 +30,7 @@ if not icon:
     icon = ":python_module"
 
 regex = re.compile(r"(\S+)(?:\s+to)\s+(\S+)")
-unitListOutput = re.compile(r"(\d+;)+[\d.]+")
+unitListOutput = re.compile(r"(\d+(e[+-]\d{2,})?;)+[\d.]+(e[+-]\d{2,})?")
 
 
 def getUnitsResult(args):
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:
don't forget to increase the version number and append your github username to the authors

iyzana:
done

ManuelSchneid3r:
>  `8 min + 20 sec`

I am not sure if I like this format. Take e.g. python timedelta. They use the format H:M:S which is well understood. Actually you could just use timedelta for readability and maintainability.

iyzana:
That could be a special case for the conversion to time.
But `units` supports arbitrary unit lists, e.g.:
```
You have: 12.28126 ft
You want: ft;in;1|8 in
        12 ft + 3 in + 3.00096 * 1|8 in

You have: 3 kg
You want: lb;oz
        6 lb + 9.8218858 oz
```

For some if these it provides a special name (like with `time`):
```
hms         hours, minutes, seconds
dms         angle: degrees, minutes, seconds
time        years, days, hours, minutes and seconds
usvol       US cooking volume: cups and smaller
ftin        feet, inches and 1/8 inches
inchfine    inches subdivided to 1/64 inch
```

Should I nevertheless implement the special case for `time` you mentioned?

ManuelSchneid3r:
bear with me I just skimmed over it sorry. I dint realize that you run the long output version. I somehow though you convert it on your own function. for whatever reason 🤷‍♂️ Thank you. LGTM

PR #98, opened 2020-12-05
[texdoc] add texdoc plugin

cit:
The texdoc plugin queries the texdoc documentation.


texdoc/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+def handleQuery(query):
+    if not query.isTriggered:
+        return
+
+    query.disableSort()
+
+    stripped_query = query.string.strip()
+
+    if stripped_query:
+        process = subprocess.run(["texdoc", "-I", "-q", "-s", "-M", stripped_query],
```
bergercookie, line 38: probably you want to use the `textdoc_cmd` that you've defined earlier

texdoc/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+from albertv0 import *
+
+__iid__          = "PythonInterface/v0.3"
+__prettyname__   = "TeXdoc"
+__version__      = "1.0"
+__trigger__      = "td"
+__author__       = "Florian Adamsky"
+__dependencies__ = ["texdoc"]
+
+iconPath   = os.path.dirname(__file__)+"/texdoc-logo.svg"
```
bergercookie, line 23: nit: I'd use `pathlib.Path(__file__)` instead of `os.path`.

texdoc/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+texdoc_cmd = ["texdoc", "-I", "-q", "-s", "-M"]
+
+if which("texdoc") is None:
+    raise Exception("'texdoc' is not in $PATH.")
+
+def handleQuery(query):
+    if not query.isTriggered:
+        return
+
+    query.disableSort()
```
bergercookie, line 33: I believe the `disableSort` method is not available in Albert v0.2 version. I'm using something like this in my plugins to the same effect:

```
# be backwards compatible with v0.2
if "disableSort" in dir(query):
    query.disableSort()
```

texdoc/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+        process = subprocess.run(["texdoc", "-I", "-q", "-s", "-M", stripped_query],
+                              stdout=subprocess.PIPE)
+        texdoc_output = process.stdout.decode("utf-8")
+
+        results = []
+        for line in texdoc_output.split("\n"):
+
+            match = re.search('\t(/.*/)([\w\.-]+)\t\t', line, re.IGNORECASE)
+            if match:
+                directory = match.group(1).strip()
```
bergercookie, line 47: Could there be  a case where < 2 groups are in the regex? If so, is it worth handling it before calling `.group(...)`

texdoc/__init__.py
dshoreman: v0.2 is incredibly old (6 years, give or take); is there really any benefit to adding complexity to support such an outdated version? If anyone *is* still using it, they're long overdue for an update ;)

texdoc/__init__.py
dshoreman: Wow, TIL! +1 for this, and after playing around in a shell it looks like it supports a fancy way to logically concat parts too.

The only downside is `pathlib.Path(__file__)` returns the file itself, so as far as I can tell we'd need to use the `parent` attribute to get the same result as `os.path.dirname()`: Even *with* that though, it's still more readable than the old method.

```suggestion
iconPath   = Path(__file__).parent / 'texdoc-logo.svg'
```

texdoc/__init__.py
```diff
@@ -0,0 +1,68 @@
+# -*- coding: utf-8 -*-
+
+"""texdoc extension
+
+This is an extension to search for LaTeX documentation.
+
+Synopsis: <trigger> <query>"""
+
+import os
```
dshoreman, line 9:
```suggestion
```

texdoc/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+import re
+import subprocess
+
```
dshoreman, line 12:
```suggestion

from pathlib import Path
```

texdoc/__init__.py
dshoreman:
```suggestion
        process = subprocess.run(texdoc_cmd + [stripped_query],
```

dshoreman (changes requested):
Unable to test run the code currently as I'm not in Linux, but I've suggested some changes should take care of the suggestion by @bergercookie to use the previously assigned `texdoc_cmd`, and `pathlib` in place of `os.path`.

Overall though, looks good! Not a huge fan of the aligned values in `results.append()`, but that's purely a nitpick.

texdoc/__init__.py
cit: Done! Thx!

texdoc/__init__.py
cit: Done! Since iconPath is not a string I need to convert it to a string when I use it. But this is not a problem.

texdoc/__init__.py
cit: I can't think of any case where the groups < 2. This matches a file and separates the directory and the filename. If texdoc has found a documentation it is always a full path with filename.

texdoc/__init__.py
bergercookie: @dshoreman haha, glad you liked it :) 

> The only downside is pathlib.Path(__file__) returns the file itself

Yup, exactly, that's why it's always more roboust IMO to use `Path(__file__).absolute().parent`.

> Path(__file__).parent

Hmm, haven't tested it, but does it work? Don't you need to use `.absolute()` before `.parent`?

bergercookie (approved):
Haven't tested the plugin itself, but the code makes sense. :+1:

texdoc/__init__.py
ManuelSchneid3r: @dshoreman bergercookie is talking of the python interface version. v0.3 is not even documented. disablesort is the only change to the interface at all. (see https://github.com/albertlauncher/plugins/blob/master/python/src/extension.cpp).

@bergercookie note that theres nothing like backwards compatibility as long as all the versions are below 1. Albert is still beta.

texdoc/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+from shutil   import which
+from pathlib  import Path
+from albertv0 import *
+
+__iid__          = 'PythonInterface/v0.3'
+__prettyname__   = 'TeXdoc'
+__version__      = '1.0'
+__trigger__      = 'td'
+__author__       = 'Florian Adamsky'
```
ManuelSchneid3r, line 20: please use your github account for easy mentions. regards

ManuelSchneid3r (changes requested):

texdoc/__init__.py
bergercookie: > @bergercookie note that theres nothing like backwards compatibility as long as all the versions are below 1. Albert is still beta.

sure, I understand, but I think that given it's a small effort to keep compatible with both, to have this sort of check before calling `.disableSort()`

texdoc/__init__.py
dshoreman: IIRC it returned the full path when called from the basedir *or* the parent dir in my testing with `python ./[path/to/]foo.py`. I also tested with `.resolve().parent` but that didn't seem to affect results. Is there much difference between `.resolve()` and `.absolute()`? I don't recall seeing the latter referenced much.

It might be worth adding `.resolve()` anyway (or absolute?) since the docs suggest it helps with symlinks, but I feel that's likely an edge case so we could probably do it in a separate PR that also updates other plugins to use `Path(...)`

texdoc/__init__.py
dshoreman: > bergercookie is talking of the python interface version. v0.3 is not even documented. 
Aha, that makes much more sense! Thanks for the link, if it hasn't been done already then I might create an issue later to get that method added in docs (unless I have time to just PR).

---

I'm on the fence here. After some digging, `disableSort` was added back in Jan 2019 (so anyone building from git will have it) but it wasn't officially released until a few weeks ago. I'm inclined to side with @ManuelSchneid3r and go the semver route, ignoring old versions since it's pre-1.0. Presumably there's a chance a lot of people are still on 0.16.1 if they're using outdated distro repositories? But in that case they probably won't have updated plugins either... Maybe? :joy: 

'scuse the essay

texdoc/__init__.py
dshoreman:
```suggestion
__author__       = 'Florian Adamsky (@cit)'
```

dshoreman (approved):
Tested locally and all working well... after an hour spent trying to get texlive/texdoc working :joy: 

Looks good!

dshoreman (approved):

ManuelSchneid3r:
Any quick tips what I have to do to test this one on archlinux?

cit:
> Any quick tips what I have to do to test this one on archlinux?

I use archlinux as well. If you have not installed LaTeX then there are two possibilities. You can either install texlive (which is recommended if you do serious LaTeX kung foo) or you can use pacman . I don't know the exact packages you need to install in order to have a minimum LaTeX environment with texdoc.  But there are ```texlive-*``` packages I would try.

https://www.tug.org/texlive/quickinstall.html
https://wiki.archlinux.org/index.php/TeX_Live

dshoreman:
@ManuelSchneid3r The quickest/easiest way is likely to install `texlive-tlpdb` from the AUR. This depends on `texlive-core` (which provides the texdoc binary) as well as all the doc packages.

When I tried to install the doc packages on their own, I would get errors saying they couldn't be found when I Tried to run `texdoc <some-package>`... even for the core texdoc package itself. The `texlive-tlpdb` package in AUR seems to solve that.

ManuelSchneid3r:
> Tested locally and all working well... after an hour spent trying to get texlive/texdoc working 
> 
> Looks good!

nah to lazy, just going to rely on you guys. ty cit.

ManuelSchneid3r (approved):

PR #100, opened 2020-12-09
base_converter: Use Python-style base prefixes to auto-detect source base 

Keating950:
Hello,

This PR streamlines the interface for the base_converter plugin a bit by automatically detecting the source base of the input by using Python's prefix rules: `0b` means a binary literal, a leading zero specifies octal, and `0x` is (of course) hexadecimal. This is a little more convenient than having to specify the input base manually, and doesn't reduce functionality given that those three bases were the only valid options to begin with. Let me know what you think.

Thanks!

### DIFF
--- a/base_converter/__init__.py
+++ b/base_converter/__init__.py
@@ -16,7 +16,7 @@ from albertv0 import Item, ClipAction
 
 __iid__ = "PythonInterface/v0.1"
 __prettyname__ = "Base Converter"
-__version__ = "2.0"
+__version__ = "1.1"
 __trigger__ = "base "
 __author__ = "Manuel Schneider"
 __dependencies__ = ["numpy"]
--- a/base_converter/__init__.py
+++ b/base_converter/__init__.py
@@ -3,26 +3,38 @@
 """Convert representations of numbers.
 
 Synopsis:
-    <trigger> <src base> <dest base> <src>
-    <bin|oct|dec|hex> <number> [padding]"""
+    <trigger> <dest base> <src>
+    <number> [padding]
+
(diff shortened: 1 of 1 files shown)


base_converter/__init__.py
```diff
@@ -3,26 +3,39 @@
...
 import numpy as np
+from collections import defaultdict
 
-from albertv0 import *
+from albertv0 import Item, ClipAction
 
 __iid__ = "PythonInterface/v0.1"
 __prettyname__ = "Base Converter"
-__version__ = "1.1"
+__version__ = "2.0"
```
ManuelSchneid3r, line 19: This is a bit messy. actually all the versions should be 0.* but well as they are they are 1.* . please use it. within each version they should be backward compatible.

ManuelSchneid3r (changes requested):

base_converter/__init__.py
Keating950: Sure, I can change it to 1.2. My rationale for 2.0 was that the syntax change is not backwards compatible. Should I change it so both the old style and the one I'm suggesting work?

ManuelSchneid3r: Thats fine, but the major version is intented to match the interface of the core. Users need no backward compatibiity for the human-machine interface :D

Keating950: Sounds reasonable enough! I've reverted it to 1.1.

ManuelSchneid3r:
trace, are there are further debugging steps you c

ManuelSchneid3r:
674fcffb76451b3025e7994b568b55e1524b1b44

PR #101, opened 2020-12-12
[remmina] Add Remmina extension

gentoo90:


remmina/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+    "RDP": iconLookup("remmina-rdp-symbolic"),
+    "VNC": iconLookup("remmina-vnc-symbolic")
+}
+
+connections = []
+
+
+class RemminaConnection:
+    def __init__(self, path):
+        global protocol_icons
```
ManuelSchneid3r, line 32: Afaik you can omit this as long as you dont write to it

remmina/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+                    try:
+                        self.icon = protocol_icons[protocol]
+                    except KeyError:
+                        pass
+
+    def __repr__(self):
+        return f"{self.name} ({self.path})"
+
+
+# Can be omitted
```
ManuelSchneid3r, line 52: drop unnecessary stuff

remmina/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+def initialize():
+    conf_dir = Path(expanduser("~/.remmina"))
+    debug(f"CONF DIR: {conf_dir}")
+    global connections
+    connections = [RemminaConnection(p) for p in conf_dir.glob("*.remmina")]
+    for c in connections:
+        info(c)
+
+
+# Can be omitted
```
ManuelSchneid3r, line 62: drop this as well if you dont use it

ManuelSchneid3r (changes requested):

ManuelSchneid3r:
@albertlauncher/python-reviewers is anybody using remmina and can test this?

bergercookie:
There's already a Remmina plugin by @okb1100 here:

https://github.com/bergercookie/awesome-albert-plugins/blob/master/plugins/remmina/__init__.py

Is there a major difference between these two?

gentoo90:
> https://github.com/bergercookie/awesome-albert-plugins/blob/master/plugins/remmina/__init__.py

:open_mouth: There's another repo full of plugins?! Maybe it would be better to add them all here? Or at least mention the repo in the documentation?

bergercookie:
> There's another repo full of plugins?! 

Yeah, I had mentioned it in the Telegram chat when I started it but the community wasn't active back then.

> Maybe it would be better to add them all here? Or at least mention the repo in the documentation?

Sure, happy to somehow integrate them to the mainline python plugins one way or another.. @ManuelSchneid3r what do you think?

remmina/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def __repr__(self):
+        return f"{self.name} ({self.path})"
+
+
+def initialize():
+    conf_dir = Path(expanduser("~/.remmina"))
+    debug(f"CONF DIR: {conf_dir}")
+    global connections
+    connections = [RemminaConnection(p) for p in conf_dir.glob("*.remmina")]
```
bergercookie, line 61: I wouldn't initialise the connections in `initialize`.
This is called only when the plugin is loaded by albert (e.g., when the user has ticked the plugin in Albert's settings.
Thus, connections added *after* the plugin has been loaded will not be detected.

Since this isn't a runtime operation, just do it  at the beginning of `handleQuery`

remmina/__init__.py
```diff
@@ -0,0 +1,83 @@
...
+    for c in connections:
+        info(c)
+
+
+def handleQuery(query):
+    pattern = query.string.strip().lower()
```
bergercookie, line 67: This function runs on every single keystroke in the albert console *even* if the `query` doesn't match the trigger.
Consider showing the connections only when `    if query.isTriggered:` is True.

bergercookie (changes requested):

bergercookie:
I've left some comments for this PR but I think it would be better to instead make a PR to the awesome-albert-plugins repo with any changes that you want to make to the existing plugin (e.g., the different icons per protocol ;) ). This way we won't have to maintain two different versions that do roughly the same thing.

dshoreman:
> Yours seems more robust, shows additional info about the connections and reads connections from `~/.local/share/remmina`.
> 
> Mine shows available connections without the trigger at the start (which I would really prefer) and reads connections from `~/.remmina`. Don't know which folder is right, seems like it saves the connections to both.

If in doubt, err on the side of the XDG dirs (`~/.config`, `~/.local/share` etc) rather than folders in the homedir which are gradually being phased out by most things.

+1 for merging those other plugins into this repo now that things are more active. For this plugin, a combination of the two would be nice - specifically being able to see the connections without a trigger (similar to how vbox VMs show)

remmina/__init__.py
gentoo90: > `if query.isTriggered:`

Aha. So that's why your version doesn't show connections without `rem` in the beginning. I'm a lazy man and would prefer less letters to type. But I see that reading all configs on each keystroke would be a waste. Can we somehow monitor file changes in a separate thread maybe?

idkCpp (changes requested):
I believe this is out of date and needs to be adjusted to the current interface.

ManuelSchneid3r:

ManuelSchneid3r:
@gentoo90 are you still interested in getting this plugin upstream?

ManuelSchneid3r:
I am going to close this PR. It is not that I dont want this to be shipped upstream, but, as is, it cant be merged due to conflicts. Users/devs get the impression that I dont want to merge plugins, probably simply due to the amount of open PRs.  It's cumbersome for me to repeatedly iterate all these abandoned and stale PRs. Note that I am still interested in merging an up to date version of this plugin and I still get notifications. So if you want something to happen leave a comment or simply open a new PR.

PR #106, opened 2020-12-29
Update __init__.py

LordBurtz:
please add my changes, furthermore I noticed that your extension is missing a __trigger__ and is not working thatfor, Ill look into this too, btw im new so sorry if I made any mistakes

### DIFF
--- a/datetime/__init__.py
+++ b/datetime/__init__.py
@@ -17,7 +17,10 @@ from albert import *
 
 __title__ = "DateTime"
 __version__ = "0.4.1"
-__authors__ = "manuelschneid3r"
+__author__ = "manuelschneid3r"  #was changed from __authors__ to __author__
+__prettyname__ = "DateTime"     #i think thats necessary too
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:
Err... You are radically reverting the API changes   why tho?

LordBurtz:
so uhm.. when I tried to use this albert module I couldnt load the it because it was missing some __ things, like author, prettyname or iid, those were required by my albert installation to even successfully load a module, so I added them. Furthermore I noticed that a trigger is missing, but Im not sure what that does with the module construct.. I think a trigger is necessary that the module works, but Im not sure (your module didnt work after adding the stuff described above and loading it successfully) thats why I wrote another time-giving module (and did a pull request). Finally: did all the github stuff work? Im fairly new to github.. sorry for any inconveniences

ManuelSchneid3r:
You have an older Albert version. Please update.

PR #107, opened 2021-01-03
Window switcher breaks looking up windows when an unknown window class exists

hackartists:
* Issue
  * If an unknown window class exists, window-switcher extension will be broken

```
$ wmctrl -l -x
0x05800008  0 N/A                   hackartist-archlinux Emulator
0x05800005  0 N/A                   hackartist-archlinux Android Emulator - femu:5554
```

* Therefore, this PR wraps `handleQuery` with error handling.
  * Practically, splitting window class raises an error.
  * However, I would say that the kind of crash can be raised on others.
  * That is why I wraps the whole code with `try-except`

### DIFF
--- a/window_switcher/__init__.py
+++ b/window_switcher/__init__.py
@@ -6,6 +6,7 @@ Synopsis: <filter>"""
 
 import subprocess
 from collections import namedtuple
+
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:

PR #110, opened 2021-02-12
Update __init__.py

djmbritt:
The current copy action does not use the standard implementation. It uses xclip, which is fine, but then it should be added as a depedency. So that users know they need to install it.


idkCpp (approved):

ManuelSchneid3r:
Why not just use the clip action then?

idkCpp:
Does the clip action support clipping the stdout of a subprocess?

ManuelSchneid3r:
Indeed not. This is somewhat subject to #983. Guess I should expose a function to set the clipboard anyway and remove the clip action.

PR #113, opened 2021-03-23
gracefully handle malformed WM_CLASS

lexh:
I noticed that Window Switcher results go missing if there is an open window with WM_CLASS name containing multiple `.` chars.

E.g. On my system, Ubuntu Software center is reported as follows

```
➜  ~ wmctrl -l -x | grep Software
0x01a00009  0 org.gnome.Software.Org.gnome.Software  quasar Ubuntu Software
```

That WM_CLASS causes the extension to barf
```
09:24:44 [warn:python.modulev1] [window_switcher] ValueError: too many values to unpack (expected 2)

At:
  /home/lex/.local/share/albert/org.albert.extension.python/modules/window_switcher/__init__.py(28): handleQuery
.
```

This patch resolves the issue, by just splitting the string once, on the final `.` char.

### DIFF
--- a/window_switcher/__init__.py
+++ b/window_switcher/__init__.py
@@ -25,7 +25,7 @@ def handleQuery(query):
             if win.desktop == "-1":
                 continue
 
-            win_instance, win_class = win.wm_class.replace(' ', '-').split('.')
+            win_instance, win_class = win.wm_class.replace(' ', '-').rsplit('.', 1)
             matches = [
                 win_instance.lower(),
                 win_class.lower(),


ManuelSchneid3r:
Check the author an mention him.

ManuelSchneid3r:
Is the rsplit intended?

lexh:
Actually I now realize that this is just weirdness on the part of the Ubuntu snap store desktop entry and is better fixed there.

ananthakumaran:
The issue happens with nautilus and probably with other applications as well. Since nautilus is a default app, I guess most of the ubuntu users would be facing this issue.

```
wmctrl -l -x | grep nautilus
0x04200028  0 org.gnome.Nautilus.Org.gnome.Nautilus xxx xxx
```

PR #114, opened 2021-04-02
[vscode]: Add new plugin for vscode projects.

corvofeng:
Signed-off-by: corvofeng <corvofeng@gmail.com>

### DIFF
--- a/vscode_projects/__init__.py
+++ b/vscode_projects/__init__.py
@@ -1,0 +1,93 @@
+# -*- coding: utf-8 -*-
+
+"""
+
+"""
+
+
+import os
+import re
+import json
+from pathlib import Path
+
+try:
+    from albert import *
+except Exception as e:
+    def info(v):
+        return print(v)
+
+    def iconLookup(v):
+        pass
+
+__title__ = "VSCode Projects"
+__version__ = "0.1.0"
+__triggers__ = "&c"
+__authors__ = "corvofeng"
+__exec_deps__ = ['code']
+
+
+storage_file = str(Path.home()) + "/.config/Code/storage.json"
+iconPath = iconLookup("visual-studio-code")
+mtime = 0
+projects = []
+
+
+def get_vscode_project(storage_json):
+    storage_dict = json.loads(storage_json)
+    entries = storage_dict['openedPathsList']['entries']
+    result = []
+    for entry in entries:
+        if 'folderUri' in entry:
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:

ManuelSchneid3r:
Hey thank you. But wait theres another yscode plugin too. Maybe you could colaborate

ManuelSchneid3r:
#132

ManuelSchneid3r:
@corvofeng are you still interested in getting this upstream? I dont know what @mparati31 and @Sharsie think about this one, but it would be cool if you guys could objectively argue which plugins do the job best or probably even better merge the best parts.

Sharsie:
@ManuelSchneid3r This implementation is kinda confusing, there's a mix of storage configs, though it's always using the sqlite under the hood. The code could use some polishing

What I can say is that the sqlite definitely contains more data to open recent paths, than the json storage. I do not know if there are any disadvantages to that though (db locking, performance), would need to be tested

Needless to say, what I would miss from both  @mparati31 and @corvofeng implementation as they are now, is the support of searching through Project Manager extension and maybe more importantly accent normalization of the query input mentioned in the [previous PR](https://github.com/albertlauncher/python/pull/132#issuecomment-1366183480). Although my implementation does include that, I would say it needs even more polishing from someone well familiar with python (which I'm not)

In my opinion, some kind of collaboration/combination would be the best solution
@mparati31 implementation is the most polished one
@corvofeng implementation is the simplest + using sqlite (which I don't know if it's a good thing or a bad thing)
Mine is the most feature rich, but needs refactoring.

All implementations could use an update to the new API as well

At this stage, I would merge neither of the three :)
I would offer to start off on some collaborated implementation, if it weren't for the fact that we are finishing a house atm and I have no time for anything until the summer. Seeing there's little progress here though, I might be done with the house earlier than anything gets merged.

Sharsie:
That is a valid point, although I would argue that it should be up to the plugin implementation whether it wants to normalize the query or not.

If you were to provide query normalization by default, you would have to provide a normalizer as well, so that the plugin developers can utilize the same implementation and don't run into mismatches where they compare the normalized query string to their own non-normalized results (e.g. this plugin searches through projects which could contain accents), or if they do normalize, their implementation could differ from the one that handles the query and thus failing the matches as well.

I'll think on this and checkout the other issue later.

PR #119, opened 2021-05-22
[baloo file] A file and directory indexer for the KDE Desktop Environment

XanderThunder:
Hello I saw that somebody requested a plugin for the widely used baloo file indexer. Since I am aswell using KDE as my DE of choice and think that this would be very useful I created such a plugin. I tested it on 2 machines of mine and the only thing which needs to be installed is the "pyxdg" dependency via `pip install pyxdg`. It uses the default breeze themes icons for the search results but this can be changed by the user as needed.

This is my first contribution to an opensource project, thus please don't expect wonders, nonetheless I tried my best.
For the basis of this plugin I heavily relied on the work of [briceio](https://github.com/briceio). I changed a few things about the search and made it work like intended with albert.


ManuelSchneid3r:

XanderThunder:
> 0.18 is out. Please check the [new api](https://github.com/albertlauncher/plugins/blob/master/python/README.md) . Do you mind to volunteer as a maintainer for the plugin?

Hi, I'm very sorry, but I'm no longer using Albertlauncher nor Linux at the moment. So I'll have to pass on the offer.
Regardless, I want to thank you for taking the time to work on this project, I enjoyed using it!

ManuelSchneid3r:
Sad. Let me know when you come back to the light side of the force :D

PR #120, opened 2021-05-25
Window switcher plus

vthuongt:
This plugin is an extended window switcher plugin for albert launcher with tokenized search and fuzzy search so that the order of search terms and exact spelling do not matter. It also is configured to only search among the windows of the current workspace/desktop. The search globally on all workspaces can be initiated by starting the search query with the character '*'.


ManuelSchneid3r:
I absolutely appreciate your time, but 

> Write programs that do one thing and do it well - Peter H. Salus

There is already a window switcher plugin. Please rather improve instead of rewriting it.

ManuelSchneid3r:

PR #127, opened 2021-11-22
Add Number Converter plugin

mparati31:


ManuelSchneid3r:
Like this https://github.com/albertlauncher/python/tree/master/base_converter?

mparati31:
Yes, very similar

ManuelSchneid3r:
I appreciate your time, but why should I merge this introducing a duplicate and confusing users, which are already annoyed by the weak UX considering plugin management?

ManuelSchneid3r:
Don't you think improving existing extensions makes more sense?

mparati31:
I agree, I didn't realize there was already an extension that did the same thing.

PR #129, opened 2022-03-16
[YouTube] updated so it works and shows thumbnails

stevenxxiu:


ManuelSchneid3r:
0.18 is out. please check the new api https://github.com/albertlauncher/plugins/blob/master/python/README.md

stevenxxiu:
I'm happy to.

Just one thing that always bugged me. Do you think it's a better idea for each plugin to be in it's own repository?

I wrote a few plugins, which do quite different things. Lumping them all in one repo make the commits rather messy.

Plugins for other projects are also in separate repos. Say for browsers, for *Sublime Text* that I use, or even for another popular launcher *Rofi*.

ManuelSchneid3r:
We discussed this a lot. Theres no clear advantage of either side. Of course it is nice to separate plugins. separation of concerns, giving credits, i even thought about interfacing the github api such that we have stars in the app as drop-in voting mechanism. This is all nice, _but_ it requires work and maintenance effort nobody is really able/willing to deliver. The core functionality of most plugins is about 20 LOC.  In my opinion having them in one place is favorable for several reasons:

* tranfsering maintainership is way easier 
* reviewing is  a lot more comfortable 
* the plugin maintainer is not a strictly the only maintainer. people can send PRs which we can merge without going through submodules. the pipeline is simply shorter. if i make systematic changes I can write sed-scripts and just commit. 
*  maintainers tend to overengineer their repos just for the before mentioned 20 LOC.
*  packagers like those on Arch will go nuts, because they have to checkout submodules in a flat structure (PKGBUILD requirement of the AUR sherrifs). this is somewhat a killer argument for me, because I have been a packager. this really sucks. 

I strongly think this list does not end here. I am quite pragmatic, therefore I prefer merging. Unless your plugin comes with hundreds of lines of code and commits each week i dont think this makes sense. the overhead is simply to huge.

stevenxxiu:
Moved to https://github.com/albertlauncher/python/pull/144.

PR #138, opened 2023-01-03
Revive CopyQ plugin

### LINKED ISSUE #137: Any plans to migrate the CopyQ plugin?
My package manager has just updated albert and I seem to be missing a few plugins. The one that I'm going to miss the most is the CopyQ plugin. 

I noticed that you archived this in 55d5f89a7bf67d5ac2cbe28f991aa5e69813f0c7

Do you have plans to add it back to the supported plugins? I presume that in it's archived state it does not support the current plugin API?

BarrensZeppelin:
Fixes #137

I've attempted to do the necessary changes to make the plugin work with the new plugin interface, and it now _works on my machine™_.

It seems like albert does not display item texts as HTML any more, so the effect of underlining the search term in the results is gone.


ManuelSchneid3r:
Yes I dropped the richtext. Do you mind to volunteer as a maintainer for this plugin?

copyq/__init__.py
```diff
@@ -0,0 +1,103 @@
+# -*- coding: utf-8 -*-
+
+"""Access CopyQ clipboard.
+
```
ManuelSchneid3r, line 4: Docsting is intended for exhaustive descritpion. DRY. There are functions for brief desc and synopsis. Rather leave it empty

copyq/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+Synopsis: <trigger> [filter]"""
+
+#  Copyright (c) 2022 Manuel Schneider
+
+import html
+import json
+import re
```
ManuelSchneid3r, line 11: Please drop all re. They are used for richtet only iirc.

copyq/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+    def id(self):
+        return __name__
+
+    def name(self):
+        return md_name
+
+    def description(self):
+        return md_description
+
+    def synopsis(self):
```
ManuelSchneid3r, line 63: I am not exactly sure but iirc default is exactly filter. May be dropped

copyq/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        return "<filter>"
+
+    def defaultTrigger(self):
+        return "cq "
+
+    def handleQuery(self, query):
+        items = []
+        q_string = query.string
+        pattern = re.compile(q_string, re.IGNORECASE)
```
ManuelSchneid3r, line 72: Re is not needed

ManuelSchneid3r (changes requested):
Looks good otherwise

copyq/__init__.py
BarrensZeppelin: Without the override of `synopsis` nothing is displayed with an empty query text.

ManuelSchneid3r: Indeed I left it empty. nmv 😅 too much to remember everything

ManuelSchneid3r (approved):

copyq/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+
+#  Copyright (c) 2022 Manuel Schneider
+
+import json
+import subprocess
+
+from albert import *
+
+md_iid = "0.5"
+md_version = "1.1"
```
ManuelSchneid3r, line 11: make it 1.2

copyq/__init__.py
```diff
@@ -0,0 +1,98 @@
+# -*- coding: utf-8 -*-
+
+#  Copyright (c) 2022 Manuel Schneider
```
ManuelSchneid3r, line 3: remove this one

copyq/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+import json
+import subprocess
+
+from albert import *
+
+md_iid = "0.5"
+md_version = "1.1"
+md_name = "CopyQ"
+md_description = "Access CopyQ clipboard"
+md_url = "https://github.com/albertlauncher/python"
```
ManuelSchneid3r, line 14: theres no md_license maybe something like bsd-2-clause or mit?

PR #140, opened 2023-01-04
Revive pass plugin

maxmil:
Take the pass python plugin from the archive and upgrade it to the API 0.5

Bump the plugin version from 0.4.1 to 0.5


pass/__init__.py
```diff
@@ -0,0 +1,107 @@
+# -*- coding: utf-8 -*-
+
+"""Manage passwords.
+
+This is a 'pass' wrapper extension.
+
+Synopsis:
```
ManuelSchneid3r, line 7: remove the docstring if it does not say more than description() and synopsis()

pass/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+    <trigger> generate <location>
+    <trigger> <filter>"""
+
+import fnmatch
+import os
+from albert import *
+
+md_iid = "0.5"
+md_version = "0.5"
+md_name = "Pass"
```
ManuelSchneid3r, line 17: please add a md_license and add yourself to md_maintainers (if you want to be a maintainer)

pass/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+                    completion="pass %s" % password,
+                    actions=[
+                        Action("copy", "Copy", lambda pwd=password: runDetachedProcess(["pass", "--clip", pwd])),
+                        Action("edit", "Edit", lambda pwd=password: runDetachedProcess(["pass", "edit", pwd])),
+                        Action("remove", "Remove", lambda pwd=password: runDetachedProcess(["pass", "rm", "--force", pwd])),
+                    ]
+                ),
+            )
+
+        return query.add(results)
```
ManuelSchneid3r, line 87: is this return intended?

ManuelSchneid3r (changes requested):

pass/__init__.py
maxmil: Removed

pass/__init__.py
maxmil: Added myself and BSD-3

pass/__init__.py
maxmil: No. Removed

pass/__init__.py
```diff
@@ -0,0 +1,101 @@
+# -*- coding: utf-8 -*-
+
+import fnmatch
+import os
+from albert import *
+
+md_iid = "0.5"
+md_version = "0.5"
```
ManuelSchneid3r, line 8: Make the version 1.2 before the interface id and plugin version has been one. 0.4 was the interface '.1' was the plugin version. Now it is iid 0.5 and version (increased) 1.2

ManuelSchneid3r:
Sorry for this iterative review

PR #144, opened 2023-01-06
feat: add *YouTube* extension

stevenxxiu:
Add the *YouTube* extension. For detailed commit messages see my repo https://github.com/stevenxxiu/albert_youtube_steven.


youtube/__init__.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+md_iid = '0.5'
+md_version = '1.0'
+md_name = 'YouTube'
+md_description = 'Query and open YouTube videos and channels.'
+md_url = 'https://github.com/stevenxxiu/albert_youtube_user'
+md_maintainers = '@stevenxxiu'
+
+ICON_PATH = str(Path(__file__).parent / 'youtube.svg')
```
ManuelSchneid3r, line 21: this is the old icon. please use the new one. you can simply reuse the one from the websearch extension

ManuelSchneid3r:
ay you hit in my rebase. can you fix that please? sorry about that

youtube/__init__.py
stevenxxiu: Sure, fixed.

stevenxxiu:
> ay you hit in my rebase. can you fix that please? sorry about that

No worries, rebased.

ManuelSchneid3r:
Does this work using the static methods? I ask because [this](https://github.com/albertlauncher/albert/blob/master/include/albert/extensions/queryhandler.h#L46-L55) class is exposed using PyBind. Since pybind does not document this I'd ask you to remove the statics. Simply because I am scared of any weird and hard to debug undefined behavior.

stevenxxiu:
It works fine with `@staticmethod`. I've never encountered any issues. I was using this since the previous *Albert Launcher* version v0.17, for quite a long time.

I just prefer using `@staticmethod` where possible, as otherwise *PyCharm* complains that there's no need to use `self`. This also makes it clear in the code which methods require `self`.

Indeed for *Pybind11* I can't find any docs regarding static methods either. I think but am not sure, that support is given by https://github.com/pybind/pybind11/pull/1732.

If it works fine, do you really see a need in removing it?

ManuelSchneid3r:
It is really not a matter of preference. Pybind exposes the internal c++ class hierarchy Extension<-Query handler. Both have pure virtual methods which _have_ to be overidden. This corresponds somewhat to ABC abstract instance methods. Maybe since Python is duck typed and you can pass kwargs, which are simply ignored then (self), this may work somehow for now, but if pybind devs decide to change this to more strictness this plugin may eventually break in future. PyCharm complains because it simply has no clue of this hierarchy.

ManuelSchneid3r:
The issue mentioned is about static defs. But as said these are no static functions. See https://github.com/albertlauncher/plugins/blob/master/python/src/plugin.cpp#L49

stevenxxiu:
Makes sense, my bad. Thanks for the explanation.

I just updated my [stevenxxiu/albert_python_stubs: PyCharm stubs for Albert Launcher Python plugins](https://github.com/stevenxxiu/albert_python_stubs) so *PyCharm* now understands this.

ManuelSchneid3r:
> I just updated my [stevenxxiu/albert_python_stubs: PyCharm stubs for Albert Launcher Python plugins](https://github.com/stevenxxiu/albert_python_stubs) so _PyCharm_ now understands this.

Oh this is really a nice to have. Unfortunately I know way to little about python for these kind of things. Do you think it is possible to ship this somehow?

youtube/__init__.py
```diff
@@ -0,0 +1,186 @@
...
+        return __name__
+
+    def name(self) -> str:
+        return md_name
+
+    def description(self) -> str:
+        return md_description
+
+    def defaultTrigger(self) -> str:
+        return 'yt'
```
ManuelSchneid3r, line 126: Give it a space suffix

stevenxxiu:
> > I just updated my [stevenxxiu/albert_python_stubs: PyCharm stubs for Albert Launcher Python plugins](https://github.com/stevenxxiu/albert_python_stubs) so _PyCharm_ now understands this.
> 
> Oh this is really a nice to have. Unfortunately I know way to little about python for these kind of things. Do you think it is possible to ship this somehow?

Yes I think this can be shipped by including a *Python* stub package. Say *Arch Linux*, can add a *Python* package to `/usr/lib/python3.10/site-packages/albert-stubs/`. I think just using my repo directly like that works, although I haven't tested.

The relevant docs are at [PEP 561 – Distributing and Packaging Type Information | peps.python.org](https://peps.python.org/pep-0561/). (I discovered a folder name issue, just fixed it.)

stevenxxiu:
Alright I updated my repo now so it can be installed directly by *Pip*, and tested that it works.

This should be able to be packaged easily now.

stevenxxiu:
Ah you didn't merge my fix to `ICON` to use the default *YouTube* icon.

ManuelSchneid3r:
By intention. :youtube is a qresource path which is only available when the websearch plugin is loaded.

ManuelSchneid3r:
But indeed to save half a kb we could just reuse it of its there

stevenxxiu:
Ok makes sense. The only other thing is I still prefer the more modern:

```python
str(Path(__file__).parent / 'youtube.svg')
```

`os.path` seems somewhat deprecated to me now.

Otherwise guess I can close this.

ManuelSchneid3r:
> `os.path` seems somewhat deprecated to me now.

it works :D i dont really care. you're the maintainer do whatever you want :D

PR #145, opened 2023-01-09
Update kill ext to 0.18

Pete-Hamlin:
PR contains:
- Updated `kill` and `currency_converter` to use 0.5 interface
- Fixed a typo in `wikipedia` so that Copy to Clipboard now works

CAVEAT - I couldn't get icons to work for love or money (still have the python symbol for all of them). Tried `xdg:` & having a local icon path, no dice. I'm more or less out of ideas on why this is happening.

As this doesn't affect behaviour, I thought I'd raise the PR and would be happy to maintain the given extensions going forward. Equally, I'm happy to close the PR I If this is an issue, and if someone else would wish to take up the mantle, they are welcome to :)


ManuelSchneid3r:
Hi Pete Jonah send a PR using python-pint which looks indeed veri promising. It more ore less replaces the gnu units and currency converter together. It is not that trivial though.

Pete-Hamlin:
Sorry for the delayed reply, this last week has been somewhat hectic!

Happy to remove currency converter and just have the PR for `kill` then if you'd prefer?

ManuelSchneid3r:
Do people actually get notified on reactions?

Pete-Hamlin:
I think it depends on your preferences, I don't appear to.

PR #146, opened 2023-01-10
Add Unit Converter

DenverCoder1:
This plugin includes a currency converter in addition to standard units supported by Pint, so the old currency converter from 0.17 is not needed if this plugin is installed.

Things to consider: 

* ~~What is the best way to handle user configuration? There is currently a `jsonc` file adjacent to the code, but I'm concerned it might be potentially overridden in this location.~~
* ~~Looks like #145 is adding a currency converter that uses the default trigger this plugin was intending to use. As mentioned above, maybe the standalone currency converter plugin there is not needed?~~


ManuelSchneid3r:
> standard units supported by Pint

is it better than the stone old gnu units?

> 

> the old currency converter from 0.17

fetches the exchangerates directly from the european central bank. Does pint have up to date values?

ManuelSchneid3r:
Looks good https://github.com/hgrecco/pint/blob/master/pint/default_en.txt

DenverCoder1:
> is it better than the stone old gnu units? 586 units, 56 prefixes

I think it's a lot easier to use. It allows multiple ways to specify units (eg. `in`, `inch`, `inches`) so it makes it more natural to use. I think everything from gnu units is supported and more.
```py
>>> len(pint.UnitRegistry()._units)
949
>>> len(pint.UnitRegistry()._prefixes)
63
```

> fetches the exchangerates directly from the european central bank

Mine does not use Pint for currency conversion, rather it fetches from https://open.er-api.com/v6/latest/USD which is updated daily.

ManuelSchneid3r:
nice! some thoughts though:

the icons are highly colorful and not really polished. Id just use a unicode beer emoji to give the users at least a funny hint and motivation to look up the lib.

please dont ship the license MIT is well known put it into the license field.

unit_converter/__init__.py
```diff
@@ -0,0 +1,460 @@
...
+`{default_trigger}88 mph to kph`
+`{default_trigger}32 degrees F to C`
+`{default_trigger}3.14159 rad to degrees`
+`{default_trigger}100 USD to EUR`
+"""
+
+md_iid = "0.5"
+md_version = "1.2"
+md_name = "Unit Converter"
+md_description = "Convert length, mass, temperature, time, currency, and more"
```
ManuelSchneid3r, line 38: Please keep this brief.  Convert units is enough. Long string make problems in the ui

unit_converter/__init__.py
```diff
@@ -0,0 +1,460 @@
...
+import pint
+
+default_trigger = "convert "
+synopsis = "<amount> <from_unit> to <to_unit>"
+
+__doc__ = f"""
+Extension for converting units of length, mass, speed, temperature, time,
+current, luminosity, printing measurements, molecular substance, currency, and more
+
+Synopsis: {default_trigger}{synopsis}
```
ManuelSchneid3r, line 24: please remove the synopsis users get it in line. thats enough.

unit_converter/__init__.py
```diff
@@ -0,0 +1,460 @@
...
+
+import albert
+import inflect
+import pint
+
+default_trigger = "convert "
+synopsis = "<amount> <from_unit> to <to_unit>"
+
+__doc__ = f"""
+Extension for converting units of length, mass, speed, temperature, time,
```
ManuelSchneid3r, line 21: it make no sense to list just _some_ units. rather generalize to pysical and financial units or just units or provide a link to a list of units. sorry to be nitpicking but trying to take the pov of a user.

unit_converter/__init__.py
```diff
@@ -0,0 +1,460 @@
...
+from datetime import datetime
+from pathlib import Path
+from typing import Any
+from urllib.request import urlopen
+
+import albert
+import inflect
+import pint
+
+default_trigger = "convert "
```
ManuelSchneid3r, line 17: try to avoid globals. they will stay in memory on unloading a plugin. if i got pybind docs correctly

unit_converter/__init__.py
```diff
@@ -0,0 +1,460 @@
...
+`{default_trigger}180 minutes to hrs`
+`{default_trigger}100 km to miles`
+`{default_trigger}88 mph to kph`
+`{default_trigger}32 degrees F to C`
+`{default_trigger}3.14159 rad to degrees`
+`{default_trigger}100 USD to EUR`
+"""
+
+md_iid = "0.5"
+md_version = "1.2"
```
ManuelSchneid3r, line 36: aint this 1.0?

unit_converter/__init__.py
```diff
@@ -0,0 +1,460 @@
...
+md_url = "https://github.com/albertlauncher/python"
+md_lib_dependencies = ["pint", "inflect"]
+md_maintainers = "@DenverCoder1"
+
+unit_convert_regex = re.compile(
+    r"(?P<from_amount>-?\d+\.?\d*)\s?(?P<from_unit>.*)\s(?:to|in)\s(?P<to_unit>.*)",
+    re.I,
+)
+
+units = pint.UnitRegistry()
```
ManuelSchneid3r, line 49: especially this. this should be part of the initialize() funtion

unit_converter/__init__.py
```diff
@@ -0,0 +1,460 @@
...
+    """
+    Strip comments and load the config from the config file.
+    """
+    with config_path.open("r") as config_file:
+        contents = config_file.read()
+    contents = re.sub(r"^\s*//.*$", "", contents, flags=re.MULTILINE)
+    return json.loads(contents)
+
+
+config_path = Path(__file__).parent / "config.jsonc"
```
ManuelSchneid3r, line 63: what exactly is in the config besides the rounding precision? Id rather drop rounding and output the most precise number we can get.

unit_converter/config.jsonc
```diff
@@ -0,0 +1,46 @@
...
+        "degree_Celsiuses": "°C",
+        "degree_Fahrenheit": "°F",
+        "degree_Fahrenheits": "°F",
+        "mile / hour": "mph",
+        "mile / hours": "mph",
+        "kilometer / hour": "km/h",
+        "kilometer / hours": "km/h",
+        "kilometer_per_hour": "km/h",
+        "kilometer_per_hours": "km/h"
+    }
```
ManuelSchneid3r, line 45: just ship these sane defaults hardcoded. honestly, nobody will ever change them.

unit_converter/__init__.py
DenverCoder1: It's coming from https://github.com/DenverCoder1/unit-converter-albert-ext which I'll probably archive when it becomes a default plugin. I can put it back to 1.0

DenverCoder1:
I've made some changes based on the feedback

unit_converter/__init__.py
ManuelSchneid3r: No okay then

Pete-Hamlin:
Author of #145 here, happy to remove currency converter from PR so this one can go ahead.

ManuelSchneid3r:
Does it need exactly inflect==6.0.2 Pint==0.20.1?

PR #151, opened 2023-01-12
feat: add *GoldenDict* extension

stevenxxiu:


ManuelSchneid3r:
If you revive old plugins please move them from the .archive before. Did this here and added some minor changes

PR #153, opened 2023-01-12
[tex_to_unicode] Migrate Tex to Unicode to 0.18 api

DenverCoder1:
Migrated tex_to_unicode and tested in 0.18

Not sure what to do about authors, license. Let me know if it looks right.

image


tyilo:
I created the original version. `GPL-3.0` as the license would be my choice of license.

ManuelSchneid3r:
Indeed thats problematic with GPL @DenverCoder1 . Since you do not hold copyright you are not allowed to changed the license. Thats the reason why I dont like GPL. But as is you have to revert to GPL or rewrite 🫤

ManuelSchneid3r:
@tyilo should I add you as a maintainer?

PR #154, opened 2023-01-13
Migrating VPN and Timer plugin to albert v18.x

uztnus:

PR #155, opened 2023-01-14
VPN (via nmcli) 1.1 and API 0.5 compatiblity

Bierchermuesli:
i forked the vpn plugin in a dedicated branch - I think this would be ready to merge


roman-acumen:
its the same as #154 right?

ManuelSchneid3r:
Yes please and call these nm, network manager or something that makes it clear that nm is used. Otherwise it will work expectation when ppl use sth else. Imho the name should represent the scope. Vpn narrows the nm scope (whatever one could imagine manipulating interfaces printing net things) while not being complete in the sense that it does not interface other von providers.

vpn/__init__.py
```diff
@@ -0,0 +1,70 @@
+"""VPN Extension
+
```
ManuelSchneid3r, line 2: this text will apear in the setings dialog. please use it only if it is beneficial for users. title is in the config and will be duplicated.

vpn/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+Connect or disconnect from a network manager VPN profile"""
+
+
+from albert import *
+from collections import namedtuple
+import subprocess
+
+md_iid = "0.5"
+md_version = "0.5"
```
ManuelSchneid3r, line 11: __version__ = "0.4.0" was iidmajor.iidminor.version there fore 1.1 makes sene i guess

vpn/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+md_id = "vpn"
```
ManuelSchneid3r, line 12: you can drop this it will automatically be set to the dir name

vpn/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+md_name = "VPN"
+md_description = "Manages VPN (via nmcli)"
```
ManuelSchneid3r, line 14: Please from the user POV "Manage VPN…

vpn/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+md_license = "MIT"
+md_url = "https://github.com/albertlauncher/python"
+md_maintainers = ["@Bierchermuesli"]
+md_authors = ["@janeklb"]
```
ManuelSchneid3r, line 18: this field does not exist

vpn/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+md_lib_dependencies = ["nmcli"]
```
ManuelSchneid3r, line 19: is this really this one https://pypi.org/project/nmcli/ ? important otherwise dep mgmt will not work

ManuelSchneid3r: wait it does not even use this package. use `md_bin_dependencies`

ManuelSchneid3r (changes requested):

PR #156, opened 2023-01-15
copyq: fix item icons by resolving item id clash

BarrensZeppelin:
I think the item id `copyq` clashes with the id of the item produced by the plugin extension for the copyq plugin.
Anyway, the copyq items were using the Python icon instead of the CopyQ icon.
Changing the id of result items seems to fix that.

### DIFF
--- a/copyq/__init__.py
+++ b/copyq/__init__.py
@@ -3,10 +3,10 @@
 import json
 import subprocess
 
-from albert import *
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:
i have to think about this a while. this is a systematic issue. sorting is a problem too and the way how items are stored in the usage database.

PR #157, opened 2023-01-26
[yt] v1.3 (iid:0.5)

stevenxxiu:
I moved the temporary directory creation from the global scope into `initialize()`, so there's not a lot of empty directories being created.

Is this a good idea? It seems that `finalize()` is called immediately after the plugin is run. Hope all icons are loaded by then.

I tested my custom plugin repo and copied most changes here. Please test to see if it still works.

### DIFF
--- a/youtube/__init__.py
+++ b/youtube/__init__.py
@@ -1,13 +1,10 @@
-# -*- coding: utf-8 -*-
-
 import json
 import re
 import tempfile
 import time
-from os import path
 from concurrent.futures import ThreadPoolExecutor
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:
> Is this a good idea? It seems that `finalize()` is called immediately after the plugin is run.

no finalize should be called before plugin instance destruction and module unloading. 

> I tested

thats fine we will see if things work out. 

generally please make sure to ship polished versions such that you dont have to ship refactorings later. imho it makes sense only to release things that actually make a difference for users.

PR #158, opened 2023-01-26
fix api_test delay option

Bierchermuesli:
- `test` as trigger makes more sense for me
- since 0.5 query.add is necessary (?)
- icon should be a list and not icon**s**

### DIFF
--- a/api_test/__init__.py
+++ b/api_test/__init__.py
@@ -26,7 +26,7 @@ md_maintainers = "@manuelschneid3r"
 
 class Plugin(QueryHandler):
     def id(self):
-        return "someid";
+        return "test";
 
     def name(self):
         return "somename";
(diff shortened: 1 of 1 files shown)


api_test/__init__.py
```diff
@@ -26,7 +26,7 @@
 
 class Plugin(QueryHandler):
     def id(self):
-        return "someid";
+        return "test";
```
tomsquest, line 29: Usually, plugin use `__name__`.
`name(self)` is `md_name`, and so on.
Do you think this test plugin should diverge from that?

Copied from other plugin:
```
    def id(self):
        return __name__

    def name(self):
        return md_name

    def description(self):
        return md_description

```

tomsquest (approved):

api_test/__init__.py
ManuelSchneid3r: Note that __name__ is albert.__name__ for plugins. That's a design decision to avoid name conflicts, e.g. that docker does not import itself. md_name is the module name if you don't override it.

ManuelSchneid3r:
Don't bother too much. This plugin is for debugging purposes only

ManuelSchneid3r:
> * since 0.5 query.add is necessary (?)

See https://github.com/albertlauncher/plugins/blob/master/python/README.md

PR #165, opened 2023-04-18
[aur] Use correct var

gaganpreet:


ManuelSchneid3r:
please increase minor version on changes

PR #169, opened 2023-05-01
[mullvad] Added mullvad vpn extension

Pete-Hamlin:
Quick little connection extension for the [mullvad VPN](https://mullvad.net/en/download/vpn/linux) app on linux.

Tested on Pop! OS 22.04, and happy to be maintainer for this one.

Figured I'd share as it's useful too me, but if it's too specifically scoped to a 3rd party app, happy to just close and leave as is.


ManuelSchneid3r:
> Figured I'd share as it's useful too me, but if it's too specifically scoped to a 3rd party app, happy to just close and leave as is.

Sure why not. But maybe it makes sense to merge all VPN plugins to a single one for improved UX since there are dozens of VPN apps (forti, wireguard, cisco etc)

Pete-Hamlin:
Not against the idea, however my main concern with that would be that each app has their own CLI interface (if they have one at all). It would make creating a standardised extension quite difficult I think.

We already have the existing `vpn` extension using nmcli, which covers OVPN/wireguard for systems than use network manager (ubuntu et al).

Not really sure how this would look as a python extension.

ManuelSchneid3r:
Hi pete,

> Not against the idea, however my main concern with that would be that each app has their own CLI interface (if they have one at all). It would make creating a standardised extension quite difficult I think.

mhm, i first thought was that it should be easy to grasp for novice users, but i think people using mullvad know that the mullvad extension is for VPN.

I'd like to merge this one but the Python api changed slightly. You could either use the master branch packages or just wait until I released 0.21 (I planned to make releases less often since some people complained about the short release cycle, thererfore the master packages)

Pete-Hamlin:
Happy to take either approach, hich is your preference?

ManuelSchneid3r:
I don't know much about the details and cant really estimate the UX. I though there is some kind of common api for VPN clients. Such that one has to choose the "backend" libraries only. but looks like e.g. has it own cli interface. so choose whatever you think achieves the best UX. if mullvad works somehow different i'd go for the dedicated plugin i guess.

Pete-Hamlin:
Sorry, had some cross contamination from my `dev` branch bleed into this PR, should be all cleaned up now.

Now version 0.21.1 is live, I think this is good to go?

ManuelSchneid3r:
sorry for the breaking changes py2.0 api is done (hopefully). please rebase. time to get some plugins merged.

ManuelSchneid3r:
I'll close unmergable PRs to reduce mental load. 😄reopen if it is ready.

PR #172, opened 2023-05-26
fix(emoji): set correct import for iid 1.0

tomsquest:
Since the change to iid 1.0, `TriggerQueryHandler` replace `QueryHandler`, which this plugin was imported.


ManuelSchneid3r:
ty

PR #173, opened 2023-06-12
feat(jetbrains_project): add `idea.sh` to the list of possible binaries

tomsquest:
The latest version of Intellij deprecated the `idea` script. 
They suggested adding `idea.sh` to $PATH.

This changes adds the `idea.sh` to the list of binaries that are searched for when the plugin starts.

### DIFF
--- a/arch_wiki/__init__.py
+++ b/arch_wiki/__init__.py
@@ -7,8 +7,8 @@ from urllib import request, parse
 import json
 import os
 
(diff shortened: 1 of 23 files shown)


ManuelSchneid3r:
sorry @tomsquest i messed up with some rebases. I added the binary.

tomsquest:
no pb. Thanks :+1:

PR #180, opened 2023-08-01
feat(pass): follow symlinks

okaestne:
Enable following symlinks to be able to use directory aliases or pass git repos added to .password-store via symlinks.

### DIFF
--- a/pass/__init__.py
+++ b/pass/__init__.py
@@ -76,7 +76,7 @@ class Plugin(PluginInstance, TriggerQueryHandler):
 
     def getPasswords(self):
         passwords = []
-        for root, dirnames, filenames in os.walk(PASS_DIR):
+        for root, dirnames, filenames in os.walk(PASS_DIR, followlinks=True):
             for filename in fnmatch.filter(filenames, "*.gpg"):
                 passwords.append(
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:
sorry for breaking things pyapi2.0 is done. please rebase

okaestne:
@ManuelSchneid3r rebased :)

ManuelSchneid3r:
@maxmil?

maxmil:
My apologies.  I've not been following development here. I can see a load of changes that are to do with  updating to the v2.0 interface which I don't know much about but I guess the question here is about this change.

https://github.com/albertlauncher/python/blob/8e07cf28af8a7d10c9fa3de2345ca967f4386293/pass/__init__.py#L79

LGTM

ManuelSchneid3r:
👍 (the new emiji paste fearture is the shit 😄)

ManuelSchneid3r:
> which I don't know much about

its not that much that changed but the changes were breaking. moved the hanlder property definition into constructors. the plugininstance is exposed to python and some addidional functions. [this](https://github.com/albertlauncher/python/blob/master/zeal/__init__.py#L13-L21) is basically what it looks like now.

PR #186, opened 2023-09-18
feat: add klipper plugin

hmnd:


ManuelSchneid3r:
or rather https://pypi.org/project/pydbus/ since 

> [dbus-python](https://dbus.freedesktop.org/doc/dbus-python/) is a legacy API, built with a deprecated dbus-glib library, and involving a lot of type-guessing (despite "explicit is better than implicit" and "resist the temptation to guess").

hmnd:
@ManuelSchneid3r looking at this again today... Perhaps [python-sdbus](https://github.com/python-sdbus/python-sdbus) might be an even better option? It seems to be the more actively maintained than the two libs you suggested.

ManuelSchneid3r:
in the end we need a small and stable library. if it is sustainable and does not pull hundreds of megabytes thats fine.

ManuelSchneid3r:
how is it going?

ManuelSchneid3r:
This issue was not intentionally but automatically closed because I renamed the master branch to main. If you are still interested in getting it merged please reopen against `main`.

PR #189, opened 2023-10-20
Bitwarden enhancements

DavidDeadly:
Primarily, I just wanted to add the 'copy_username' action because it's something I've been needing regularly these days. In the proccess of understanding I found a few things that could be improved, and I just did it. 🤓 The commits have explanations of my decisions.

Initially, I considered keeping this changes local, but I then realized  was the perfect time to help and do my first ever open source contribution. Hope to be helpful!

I'm totally open to debate and feedback, thanks!

### DIFF
--- a/bitwarden/__init__.py
+++ b/bitwarden/__init__.py
@@ -20,7 +20,7 @@ class Plugin(PluginInstance, TriggerQueryHandler):
 
     def __init__(self):
         TriggerQueryHandler.__init__(self,
-                                     id=md_iid,
+                                     id=md_id,
                                      name=md_name,
                                      description=md_description,
                                      defaultTrigger='bw ')
--- a/bitwarden/__init__.py
+++ b/bitwarden/__init__.py
@@ -27,7 +27,7 @@ class Plugin(PluginInstance, TriggerQueryHandler):
         PluginInstance.__init__(self, extensions=[self])
         self.iconUrls = [f"file:{Path(__file__).parent}/bw.svg"]
 
-    def _get_passwords(self):
+    def __get_items(self):
         field_names = ["id", "name", "user", "folder"]
         p = run(
             ["rbw", "list", "--fields", ",".join(field_names)],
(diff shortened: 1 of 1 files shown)


tomsquest (commented):
:+1: for the contribution!

Minor styling issue: the "double underscore" could be replace with single underscore, which is traditionally used for private members.
> [pep8] _single_leading_underscore: weak "internal use" indicator. E.g. from M import * does not import objects whose name starts with an underscore.

DavidDeadly:
Sure!! I was kind of confused about that, but now I know. Thanks!!

ManuelSchneid3r:
@ovitor @tylio what do you think?

bitwarden/__init__.py
```diff
@@ -20,86 +20,35 @@ class Plugin(PluginInstance, TriggerQueryHandler):
 
     def __init__(self):
         TriggerQueryHandler.__init__(self,
-                                     id=md_id,
+                                     id=md_iid,
```
ManuelSchneid3r, line 23: dont use md_iid here. md_id is the id of the plugin, md_iid is the interface id, currently[ v2.1](https://raw.githubusercontent.com/albertlauncher/python/master/albert.pyi)

DavidDeadly: Sorry, I thought it was a misspelling because I didn't find where 'md_id' was declared, the linter marked as not defined

ManuelSchneid3r:
please also increase the version.

ManuelSchneid3r:
Actually I meant the plugin version not the interface id. To me it looks like the plugin has some new features. If the UX changes make it 2.0 if it's just some non user recogniceable changes make it 1.4.

DavidDeadly:
Right, my bad... 

Well, I added an action (copy username), that's the only recognisable change in what respects to the user, not sure if that's enough to make a major, please confirm it.

And should I leave the interface id in v2.1?? It's the currently active, isn't it?

ManuelSchneid3r:
Go with 2.0. I dont remember the additions to 2.1 exacty and forgot to document them. Yeah better leave 2.1 in there

PR #190, opened 2023-11-06
fix [unit_converter]: Update unit_converter to API v2.1

Pete-Hamlin:
Updated the `unit_converter` python extension to work with API 2.1. Also added support for a `GlobalQueryHandler` via the settings.

@DenverCoder1 I added myself as a maintainer as I'm happy to keep an eye on this one in case of breaking API changes. As this is your work though I'm happy to remove myself if you would prefer it.

PR #195, opened 2023-12-17
Resurrecting the unit_converter

ismasou:
This is just a quick update to unit_converter to make it work with the new api.
It's working for me for a while now, but let me know if you find some problem.


ManuelSchneid3r:
#190 as discussed

PR #196, opened 2024-02-16
[vscode_projects:1.0] Release plugin

Sharsie:
I have done a rewrite of my previous implementation into the new API, I know it is a more complex plugin than the other options, but my long standing goal is to support the Project Manager extension without relying too much on the recently opened files.

Additionally this implementation provides optional configuration to open VS Code through terminal for advanced use, e.g. where users need to load an environment prior to opening VS Code for example.
My usecase is direnv + nix shell, I have separate environments in each of my projects and entering those directories in shell loads all necessary dependencies and binaries needed for that specific project. This also requires VS Code to be opened from that directory in order to have the correct environment inside the IDE.

So here it is, if you find it useful, I'm happy to support it going forward, if not, I'll just support myself :)

### DIFF
--- a/vscode_projects/__init__.py
+++ b/vscode_projects/__init__.py
@@ -1,0 +1,542 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2024 Sharsie
+
+"""
+Recent files are sorted in order found in the VSCode configuration.
+
+Sort order with Project Manager can be adjusted, lower number = higher priority = displays first.
+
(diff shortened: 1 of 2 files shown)


Sharsie:
@ManuelSchneid3r I assume all the open PRs got closed due to master->main branch rename? Should we reopen against main?

ManuelSchneid3r:
Oh, I thougt github would be smart enough to see that main actually points to the same commit as master. Sure please reopen.

PR #123, opened 2021-07-31
Add vscode projects plugin as a submodule

Sharsie:
Following in the steps of @mqus , this PR will add Visual Studio Code autosuggestion support as a submodule to albert launcher python plugins.

Since I was meddling with the jetbrains plugin in the past, I took the liberty to make something similar for VS Code.

Supports
- autosuggestions for recently opened paths as seen in the VS Code menu dialog - builds on top of the more verbose menu entries
- Projects managed by [Project Manager](https://marketplace.visualstudio.com/items?itemName=alefragnani.project-manager) plugin

This is an extension to https://github.com/albertlauncher/python/pull/114

PR #198, opened 2024-03-05
[vscode_projects:1.0] Release plugin

Sharsie:
I have done a rewrite of my previous implementation into the new API, I know it is a more complex plugin than the other options, but my long standing goal is to support the Project Manager extension without relying too much on the recently opened files.

Additionally this implementation provides optional configuration to open VS Code through terminal for advanced use, e.g. where users need to load an environment prior to opening VS Code for example.
My usecase is direnv + nix shell, I have separate environments in each of my projects and entering those directories in shell loads all necessary dependencies and binaries needed for that specific project. This also requires VS Code to be opened from that directory in order to have the correct environment inside the IDE.

So here it is, if you find it useful, I'm happy to support it going forward, if not, I'll just support myself :)

### DIFF
--- a/vscode_projects/__init__.py
+++ b/vscode_projects/__init__.py
@@ -1,0 +1,542 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2024 Sharsie
+
+"""
+Recent files are sorted in order found in the VSCode configuration.
+
+Sort order with Project Manager can be adjusted, lower number = higher priority = displays first.
+
+With all priorities equal, PM results will take precedence over recents.
+
+PM extension: https://marketplace.visualstudio.com/items?itemName=alefragnani.project-manager
+
+Terminal Command:
+
+You can override how VSCode is opened.
+Terminal will enter the working directory of the project upon selection and execute the desired command.
+
+E.g. If you are using different environments based on the working dir, such as direnv, you can open
+vscode by first executing direnv to load your environment by providing the following terminal command:
+
+direnv exec . code .
+"""
+
+import os
+import json
+import unicodedata
+from pathlib import Path
+from dataclasses import dataclass
+from albert import *
+
+md_iid = "2.1"
+md_version = "1.0"
+md_name = "VSCode projects"
+md_description = "Open VSCode projects"
+md_url = "https://github.com/albertlauncher/python/tree/master/vscode_projects"
+md_license = "MIT"
+md_bin_dependencies = ["code"]
+md_maintainers = "@Sharsie"
+
+@dataclass
+class Project:
+    displayName: str
+    name: str
+    path: str
+    tags: list[str]
+
+
+@dataclass
+class SearchResult:
+    project: Project
+    # priority is used to sort returned results
+    priority: int
+    # sortIndex is a decision maker when two search results have same priority
+    sortIndex: int
+
+
+@dataclass
+class CachedConfig:
+    projects: list[Project]
+    mTime: float
+
+
+class Plugin(PluginInstance, TriggerQueryHandler):
+    # Possible locations for Code configuration
+    _configStoragePaths = [
+        os.path.join(os.environ["HOME"], ".config/Code/storage.json"),
+        os.path.join(os.environ["HOME"],
+                     ".config/Code/User/globalStorage/storage.json"),
+    ]
+
+    # Possible locations for Project Manager extension configuration
+    _configProjectManagerPaths = [
+        os.path.join(
+            os.environ["HOME"], ".config/Code/User/globalStorage/alefragnani.project-manager/projects.json")
+    ]
+
+    # Indicates whether results from the Recent list in VSCode should be searched
+    _recentEnabled = True
+
+    # Indicates whether projects from Project Manager extension should be searched
+    _projectManagerEnabled = False
+
+    # Defines sorting priorities for results
+    _sortPriority = {
+        "PMName": 1,
+        "PMPath": 5,
+        "PMTag": 10,
+        "Recent": 15
+    }
+
+    # Holds cached data from the json configurations
+    _configCache: dict[str, CachedConfig] = {}
+
+    # Overrides the command to open projects
+    _terminalCommand = ""
+
+    # Setting indicating whether results from the Recent list in VSCode should be searched
+    @property
+    def recentEnabled(self):
+        return self._recentEnabled
+
+    @recentEnabled.setter
+    def recentEnabled(self, value):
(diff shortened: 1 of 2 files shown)


ManuelSchneid3r:
Hi, quite a bunch of lines but i guess people will love it. Just a note: py api v2.3 is out. theres a matcher which does imlicit normalization and such.

Sharsie:
I updated the plugin to use the latest version and the matcher, but noticed the iconUrls defined on the plugin itself stopped working with 0.24.1 (this is not just this plugin, but all of them). Icons on the item results work fine.

Are there any notes on how to migrate these? I noticed in the c++ version of the Timer plugin, that this is provided through a [.qrc file](https://github.com/albertlauncher/plugins/blob/main/timer/timer.qrc)

vscode_projects/__init__.py
```diff
@@ -0,0 +1,545 @@
...
+                callable=lambda: runDetachedProcess(
+                    ["code", project.path]),
+            )
+        )
+
+        return StandardItem(
+            id=project.path,
+            text=project.displayName,
+            subtext=project.path,
+            iconUrls=[f"file://{Path(__file__).parent}/icon.svg"],
```
ManuelSchneid3r, line 364: what if you remove the two slashes?

ManuelSchneid3r:
> Are there any notes on how to migrate these? I noticed in the c++ version of the Timer plugin, that this is provided through a [.qrc file](https://github.com/albertlauncher/plugins/blob/main/timer/timer.qrc)

thats the c++ way to ship a plugin as a single module. icon urls should actually work as expected. they work on my system. but yes I changed the icon provider api and maybe introduced a bug. i have to check it on a linux box.

vscode_projects/__init__.py
Sharsie: Leaving it there was an oversight, will push the commit momentarily, I was actually testing multiple cases and in my case `file:...` works, `file://...` works as well and suprisingly any amount of slashes works too.
I was refering to the [initialization](https://github.com/albertlauncher/python/pull/198/files#diff-3c8b5db043c3c17ea3751645c9353138203d2aa594c8a67eafdf994ee47b381bR163), that's where it stopped working. It stopped working for all python plugins, it shows some kind of fallback symbol instead:

image

Sharsie: Just to clarify, it works fine for the StandardItem class:
image

ManuelSchneid3r: The mpris icon problem looks like a font config issue. The icons there use emojis.

ManuelSchneid3r: IMAGE 2024-06-30 13:09:28

this is what it usually looks like (this is on vanilla manjaro)

Sharsie: Ah okay, that makes sense, so where would the `self.iconUrls` be shown? Maybe a better example is the [iconUrls in wikipedia plugin](https://github.com/albertlauncher/python/blob/main/wikipedia/__init__.py#L28)

When I enter `wiki`, I get the following results:

image

I would assume the first icon would use the iconUrls, but seems to fallback to some kind of emoji that my system font cannot render (my problem, not blaming anyone:)
The second one I would understand uses the "puzzle" emoji (just like mpris configuration row from your screenshot)

Though when I enter `wiki something` I get the iconUrl, but that is already rendered through StandardItem result

image

ManuelSchneid3r: These items are not from the wiki query handler. The first one is the built in TriggersQueryHandler and the second from the pluginqueryhandler

Sharsie: Understood, so is there any point specifying the iconUrls on the plugin itself? Searching through the source code, I don't really see many references to that except for FallbacksModel, but I'm not exactly sure what that is used for. But majority of python plugins still define it on the plugin instance.

ManuelSchneid3r: Yes, to save reasources and fast query times. The icons stored in the plugin instance are static, so there is no need to build a path and list on every query for each item.

Sharsie: > The icons stored in the plugin instance are static, so there is no need to build a path and list on every query for each item

Understood, I'll make the changes as appropriate, my mistake

> Do you struggle to understand the API? If so do you know the stubfile (albert.pyi)?

I think I just did not connect the dots, I barely use python and the whole concept of decorators for properties and such is a little bit foreign to me, so originally (a few years ago) I just modified the jetbrains plugin for vscode. Nowdays I'm a little bit more familiar with both the language and the API, but the I totally missed the path caching for some reason.

> I dont get that. I dont think the fallbacksmodel uses plugininstance fields.

No, that was due to my misunderstanding of iconUrls usage on the plugin instance. I was trying to find references where it could be used in the albert core. Now it's clear to me that it's just internally used inside the plugins themselves.

Sorry for the confusion and thanks for the patience :)

ManuelSchneid3r:
@Sharsie you can drop the string normalization the mather does this already unless configured otherwise. [see this](https://albertlauncher.github.io/reference/classalbert_1_1MatchConfig.html)

Sharsie:
Oh, very cool, dropped

ManuelSchneid3r:
import unicodedata is still there

Sharsie:
damn, that's what happens when you do a quick edit in terminal :) Thanks for catching it, fixed

Silureth:
@Sharsie thanks for returning! Been using your extension for a long time with early versions of Albert. Just fixing/clobbering it enough to work with the current versions (have my own Frankenstein repo) but since I'm not too comfortable in python never tried to added it to the official plugins.

Sharsie:
Wow, merged after [4 years](https://github.com/albertlauncher/python/pull/123). Yea I get the frustration, maybe it's just too niche.

I'll support it to the best of my ability, will probably make some updates to support vscode forks too, since nowdays I keep another version locally.

@Silureth I've been trying to keep it up to date for years now, but feel free to ping me if something breaks.

ManuelSchneid3r:
I am sorry but I don't use it. As such I am not really a suitable reviewer. Generally I don't merge unreviewed code and don't review code I can't understand.

I just wonder it is not that niche why not a single reviewer took the effort.

Sharsie:
1st - Oh dear, I fixed that in my other local plugin and forgot to update this one. If you don't mind, send a PR please.

2nd - Thanks, I'll take a look, there are occasional changes and for one of the vscode forks, this actually stopped working recently and recents can only be queried through sqlite instead of parsing the config file. I'll check if that's the case and update it

PR #199, opened 2024-03-06
[local_bookmark:1.0] feat: Implement MVP

MortalHappiness:
# Albert Local Bookmark Plugin

Open the URLs defined in the local plain text file.

## The format of the bookmark file

```
# Keyword 1
http://example1.com
# Keyword 2, Keyword 3
http://example2.com
# keyword 4
http://a.example3.com
http://b.example3.com
# Keyword 5
http://a.example4.com
# Keyword 5
http://b.example4.com
```

That is:
- The keywords are preceded by a `#` and are comma-separated, and the URLs are listed below the keywords.
- Empty lines are ignored.
- The keyword and URL mappings can be 1 to 1, 1 to many, or many to 1.
- URLs with the same keyword are allowed, as the "Keyword 5" shown in the example above.
- Keywords are case-insensitive.

### DIFF
--- a/local_bookmark/README.md
+++ b/local_bookmark/README.md
@@ -1,0 +1,30 @@
+# Albert Local Bookmark Plugin
+
+Open the URLs defined in the local plain text file.
+
+## The format of the bookmark file
+
+```
+# Keyword 1
+http://example1.com
+
+# Keyword 2, Keyword 3
+http://example2.com
+
+# keyword 4
+http://a.example3.com
+http://b.example3.com
+
+# Keyword 5
+http://a.example4.com
+
+# Keyword 5
+http://b.example4.com
+```
+
+That is:
+- The keywords are preceded by a `#` and are comma-separated, and the URLs are listed below the keywords.
+- Empty lines are ignored.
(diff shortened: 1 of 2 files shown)


ManuelSchneid3r:
To be honest (while trying to be not offensive) the concept of this plugin and the quality of its implementation does not convice me. Some points:

- The need. We have bookmark plugins (well unfortunately only chromium. thats a good point to start). Why though?
- Python is a script language. This is a lot of code just to do the equivalent of having a dict the user could modify. Eg like
  ```python
  d = {
  'http://my.url': ['s1', 's2', 's3'], 
  …
  }
  ```
- The file is read and parsed and matched __for every query__
- Does not use index tooling provided by the core. instead of using the power off C++ it pulls dependencies to achieve the same in python.

Even if the technical aspects would not apply I would rather not merge it because this is an absolute niche case. Hardly anybody will use it. So I dont think it makes sense to ship it upstream.

ManuelSchneid3r:
You can still put the project on awesome_albert. Where people can find and use it. Also there is some sort of python hub planned. If we have it I could easily ping you.

MortalHappiness:
@ManuelSchneid3r OK. Thanks for your recommendation!

PR #201, opened 2024-07-11
Feat: Better Python Eval

rtk-rnjn:
## Features:
  - `print(...)` prints the object, instead of returning `None`
  - Displays Execution time instead of `str`
  - Handles `return`, `yield`, and `await ...` statements/expressions

## Screenshots (testing):
  Screenshot from 2024-07-11 11-59-31
  Screenshot from 2024-07-11 12-00-38
  Screenshot from 2024-07-11 12-01-11

### DIFF
--- a/python_eval/__init__.py
+++ b/python_eval/__init__.py
@@ -65,7 +65,6 @@ async def _coro_repl():
 
 
 class KeywordTransformer(ast.NodeTransformer):
-    # Author: Gorialis
     # Source: https://github.com/Gorialis/jishaku/blob/master/jishaku/repl/walkers.py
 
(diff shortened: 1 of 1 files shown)


rtk-rnjn:
Also handles complex expr, like
image

ManuelSchneid3r:
"better" is subjective. Tbh I think none of these are necessary. This is a quick python evaluator. Why would I use print or await. Also time is less important to me than the type. This also introduces more code to maintain for such a basic plugin

ManuelSchneid3r:
Sorry I won't merge this upstream. this is way too much code to maintain for me. You can still put your repo in the awesome-albert repo. Regards

PR #202, opened 2024-08-05
Better search jetbrains and add new applications

d3v2a:
Add the possibility to search with project name (.iml files)

Add Aqua and Writerside to compatible editor

### DIFF
--- a/jetbrains_projects/__init__.py
+++ b/jetbrains_projects/__init__.py
@@ -125,6 +125,11 @@ class Plugin(PluginInstance, TriggerQueryHandler):
                 binaries=["studio", "androidstudio", "android-studio", "android-studio-canary", "jdk-android-studio",
                           "android-studio-system-jdk"]),
             Editor(
+                name="Aqua",
+                icon=plugin_dir / "icons" / "aqua.svg",
+                config_dir_prefix="JetBrains/Aqua",
+                binaries=["aqua", "aqua-eap"]),
+            Editor(
                 name="CLion",
                 icon=plugin_dir / "icons" / "clion.svg",
                 config_dir_prefix="JetBrains/CLion",
@@ -181,6 +186,11 @@ class Plugin(PluginInstance, TriggerQueryHandler):
                 icon=plugin_dir / "icons" / "rustrover.svg",
                 config_dir_prefix="JetBrains/RustRover",
                 binaries=["rustrover", "rustrover-eap"]),
+            Editor(
+                name="Writerside",
+                icon=plugin_dir / "icons" / "writerside.svg",
+                config_dir_prefix="JetBrains/Writerside",
+                binaries=["writerside", "writerside-eap"]),
         ]
         self.editors = [e for e in editors if e.binary is not None]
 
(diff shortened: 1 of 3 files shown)


jetbrains_projects/__init__.py
```diff
@@ -84,15 +84,22 @@ def _parse_recent_projects(self, recent_projects_file: Path) -> list[Project]:
...
+                project_path = entry.attrib["key"]
+                files = Path(project_path + "/.idea").glob("*.iml")
+                project_path = project_path.replace("$USER_HOME$", str(Path.home()))
+                project_name = Path(project_path).name
                 tag_opened = entry.find(".//option[@name='projectOpenTimestamp']")
                 last_opened = tag_opened.attrib["value"] if tag_opened is not None and "value" in tag_opened.attrib else None
 
                 if project_path and last_opened:
                     projects.append(
                         Project(name=Path(project_path).name, path=project_path, last_opened=int(last_opened))
```
tomsquest, line 96: refactor: you can also use `project_name` here instead of  `Path(project_path).name`

jetbrains_projects/__init__.py
```diff
@@ -84,15 +84,22 @@ def _parse_recent_projects(self, recent_projects_file: Path) -> list[Project]:
 
             projects = []
             for entry in entries:
-                project_path = entry.attrib["key"].replace("$USER_HOME$", str(Path.home()))
-
+                project_path = entry.attrib["key"]
+                files = Path(project_path + "/.idea").glob("*.iml")
+                project_path = project_path.replace("$USER_HOME$", str(Path.home()))
+                project_name = Path(project_path).name
```
tomsquest, line 90: style: can you group the variables like the following?

```
                project_path = entry.attrib["key"]
                project_path = project_path.replace("$USER_HOME$", str(Path.home()))
                project_name = Path(project_path).name
                files = Path(project_path + "/.idea").glob("*.iml")
```

tomsquest:
Hi @d3v2a ,

Question: 
If I get it right, the code use the name of the `.iml` file to define projects.
For example, one could have a project `/myprojects/projectFoo/.idea/bar.iml` and the code will add `bar` as an entry.
Is that right ?
(note that the current code, also add `projectFoo` as an entry)

d3v2a:
@tomsquest 
Yes, when you rename project (File ⇾ rename project), it creates an iml file with the new name

tomsquest (approved):

tomsquest:
Good to merge @ManuelSchneid3r

tomsquest:
Thanks for your changes @d3v2a !

ManuelSchneid3r:
please increase the version to 2.0

ManuelSchneid3r:
i have a problem with this one:

```python
projects = [p for p in projects if query_lower in p.name.lower() or query_lower in p.path.lower()]
```


Bildschirmfoto 2025-03-26 um 10 42 09

is this path search necessary for you guys?

tomsquest:
Hi @ManuelSchneid3r ,

> is this path search necessary for you guys?

In my case, the path is not needed, only the project's name.

:+1: to remove it.

d3v2a:
HI @tomsquest @ManuelSchneid3r,

We can add a search in path option without removing this.

ManuelSchneid3r:
How should i name the option label? 'Match path'?

PR #204, opened 2024-09-27
[window_switcher:0.5.0]

nopsqi:


window_switcher/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+
+Window = namedtuple("Window", ["wid", "desktop", "wm_class", "host", "wm_name"])
+
+
+class Plugin(PluginInstance, TriggerQueryHandler):
+    def __init__(self):
+        PluginInstance.__init__(self)
+        TriggerQueryHandler.__init__(
+            self, self.id, self.name, self.description,
+            defaultTrigger='w '
```
ManuelSchneid3r, line 30: iirc in 3.0 this will be a function

ManuelSchneid3r (changes requested):
other things: 
- rename it to x_window_switcher
- check for an x session in the constructor an raise if something else

PR #207, opened 2024-11-12
[jetbrains_projects] Fix Rider config

albi005:
Fixes https://github.com/albertlauncher/albert/issues/1473

### DIFF
--- a/jetbrains_projects/__init__.py
+++ b/jetbrains_projects/__init__.py
@@ -56,11 +56,21 @@ class Editor:
     config_dir_prefix: str
     binary: str
 
-    def __init__(self, name: str, icon: Path, config_dir_prefix: str, binaries: list[str]):
+    # Rider calls recentProjects.xml -> recentSolutions.xml and in it RecentProjectsManager -> RiderRecentProjectsManager
+    is_rider: bool
+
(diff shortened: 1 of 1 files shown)


tomsquest (approved):
Ah Jetbrains... :/

I would have skipped `is_rider` and detect it inline in the Editor class, but LGTM.

danny-isc:
Is there anything blocking this PR from being merged?

ManuelSchneid3r:
At least the fact that api is not compatible. But that's a minor issue. Generally I think there is room for improvement here, but I am busy with core things and actually it works. So for me it is low priority.

albi005:
What do you mean by "API not compatible"?

Should I remove the `is_rider` field? I felt like being a bit more explicit was better here, and if there are more edge cases, it can be refactored out. Until then, I think it's fine.

ManuelSchneid3r:
> What do you mean by "API not compatible"?

https://github.com/albertlauncher/plugins/blob/main/python/albert.pyi

0.27 came with a python api update

ManuelSchneid3r:
Ah wait. I clicked view file above and saw the old file. But probably the merge only merges the changes not the entire file.

PR #208, opened 2025-01-11
Add pycharm-professional to binary list

mqus:
Hi! I noticed that pycharm projects were not popping up for me, so this seems to be the fix. For reference, this is happening on archlinux with the [pycharm-professional](https://aur.archlinux.org/packages/pycharm-professional) package from AUR.

Thank you for maintaining the plugin!


tomsquest (approved):

tomsquest:
:+1: 

I don't know what is going with all the different naming, but that's good anyway

PR #210, opened 2025-02-20
GitHub repositories search plugin

AirOnSkin:


github/__init__.py
```diff
@@ -0,0 +1,192 @@
...
+Rebuild cache with 'gh rebuild cache'
+"""
+
+import os
+import json
+import keyring
+from albert import *
+from pathlib import Path
+import github3
+from rapidfuzz import fuzz
```
ManuelSchneid3r, line 16: albert provides a builtin matcher. pleease use it. do not search fuzzy by default. see an example https://github.com/albertlauncher/python/blob/53c4bdda1c4edcef226159deef4279739414fda0/jetbrains_projects/__init__.py#L229-L241

AirOnSkin: Done. I took the code from the emoji plugin as an example instead to make the fuzzy search an option.

tchiotludo:
@ManuelSchneid3r warm me about this PR I was not aware. In a meantime, I've created an alternative vision for this plugin: https://github.com/albertlauncher/python/pull/214 
Do you want to me to add some commit that would add the issue / pr search also ? 

Btw, I like the idea that you have some local cache for the repository list. 
Tell me what you prefer ?

AirOnSkin:
> @ManuelSchneid3r warm me about this PR I was not aware. In a meantime, I've created an alternative vision for this plugin: #214 Do you want to me to add some commit that would add the issue / pr search also ?
> 
> Btw, I like the idea that you have some local cache for the repository list. Tell me what you prefer ?

Adding PR & issue search would be lovely 👍🏻 
I would prefer to stick with `github3.py` instead of `PyGithub`. The reason being that PyGithub doesn't support Python 3.13 (yet).
The way the GitHub token is persisted can be discussed. If storing it as part of the Albert config is as secure as within a keyring I'd be fine with that, too.

AirOnSkin:
@tchiotludo I'd like to merge the PR at some point. Will you want to add your changes?

tchiotludo:
I don't have to the time to port my issue search to your version, but I would love to do it (just I didn't have the time for now). 

Your way to manage the auth token is better than mine I think, mine let the token in clear text inside the configuration. 

Maybe you could merge your work and I will try to update it including the issue / pull request search, it's just a prefix: 

```
gh r test # repos search 
gh i test # issues search 
gh p test # pull requests search 
```

With a configuration to choose the default one? 

WDYT?

ManuelSchneid3r:
Given the architecture of albert a keychain is not really more secure, is it? 

an attacker that has access to the config file could also add and enable a plugin that reads the keychain on next start. afaict the only real security would probably be signing and subprocessing of plugins such that the user has fine grained control over the keychain access. 

not sure though. I am not a security expert.

AirOnSkin:
> I don't have to the time to port my issue search to your version, but I would love to do it (just I didn't have the time for now).
> 
> Your way to manage the auth token is better than mine I think, mine let the token in clear text inside the configuration.
> 
> Maybe you could merge your work and I will try to update it including the issue / pull request search, it's just a prefix:
> 
> ```
> gh r test # repos search 
> gh i test # issues search 
> gh p test # pull requests search 
> ```
> 
> With a configuration to choose the default one?
> 
> WDYT?

Yeah, that sounds good to me 👍🏻
I like the idea of having a default target with the option to switch to others.
Following the discussion in #214, I second the ability for users to define their own shortcuts.

ManuelSchneid3r:
If you have any ideas let me know: https://github.com/albertlauncher/albert-plugin-github/issues/1

PR #212, opened 2025-03-07
Entries caching for Bitwarden plugin

DavidDeadly:
This implements what was requested on [this discussion](https://github.com/orgs/albertlauncher/discussions/1512)

The output of `rbw list` will be cached, by default, for 5 minutes or until sync action is called. This time is configurable 

image

### DIFF
--- a/bitwarden/__init__.py
+++ b/bitwarden/__init__.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 
 import time
+from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:
Ain't there a way to detect changes?

DavidDeadly:
Do you mean, somehow detect when an entry has changed or updated?
I can't think right now in an easy way to do it
Maybe run in the background `rbw list`?

PR #213, opened 2025-03-11
PWGen

LorenzoMorelli:
Extension useful to generate random passwords of different length.


Silureth (approved):
Tried it out and works, can't see any problem with the code itself.
Some extra options would be nice through configuration (length, included chars bu those are extra features)

ManuelSchneid3r:
Please see https://github.com/albertlauncher/python/issues/226

PR #214, opened 2025-03-30
[github] introduce github plugins

tchiotludo:
First plugin, tell what do you think


ManuelSchneid3r:
Theres https://github.com/albertlauncher/python/pull/210. Please colaborate to get best of both worlds.

tchiotludo:
oops I didn't see this one, even it's seems closed since related to github, my plugin seems to target wider area of the github api (issues, pull request and repositories) and that can be extended (based on the search prefix). I will asked to @AirOnSkin if he want to me to wider it's plugins

ManuelSchneid3r:
I wrote a cpp version because i needed a playground for the internal tooling (oauth, rest). 
https://github.com/albertlauncher/albert-plugin-github/blob/main/src/plugin.cpp
Id like to discuss the interface. `u`, `r`, `i` are there. However they are mostly useless. currently there are two global items that map to the user specify search of repos and issuses. however this is not that convenient and rigid. I thought about adding `R` and `I` which use default search parameters. this however is less transparent and may confuse users. 
Another approach is to let users define their own shotcuts.

tchiotludo:
Nice work :clap: 

I would agree that the `r` `u` `i` prefix are not optimal but I didn't know how to manage this in the Albert landscape (especially in python) 

The best UX IMO could be to allow people configure multiple triggers (`ghi`, `ghr`) with a  configuration for each one. 
This will allow crazy things like: 
- `ghme`: `state:open assignee:@me sort:updated-desc`
- `ghr`: list of repositories
- `ghorg`: state:open org:albertlauncher sort:updated-desc`
- `ghpr`: `is:pr state:open org:albertlauncher sort:updated-desc`

We could clone the WebSearch plugin configuration to achieve this configuration 
image

with : 
- **type:** `ISSUES` `PRS` `REPOSITORIES` `USERS` 
- **filters:**: the search query 
- **shortcut**: the shortcut used 

The last remaining thing how to get the personal access token  

WDYT?

AirOnSkin:
I like the idea of user defined shortcuts a lot. If the CPP version of the plugin will be included in Albert in the future my PR #210 can be closed.

AirOnSkin:
> Any ideas for improvement?

I'd say go ahead and release it. Improvement requests will come when people start using it.

ManuelSchneid3r:
hey guys please do not put more effort in this. i am currently implementing this in a native plugin.
If you have any ideas let me know: https://github.com/albertlauncher/albert-plugin-github/issues/1

PR #219, opened 2025-06-19
Update bw plugin to not crash and extend functionality

abbasou:
This is a modified (with the help of AI) version of the Albert BW plugin that works fast, displays *all* fields, including custom fields, in the "alt" menu while hiding sensitive fields, and doesn't crash. I'm not a developer (hence the AI), but this works flawlessly on my self-hosted instance of vaultwarden. Cheers.

### DIFF
--- a/bitwarden/__init__.py
+++ b/bitwarden/__init__.py
@@ -1,74 +1,70 @@
 # -*- coding: utf-8 -*-
 
-import time
-from dataclasses import dataclass
-from enum import Enum
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


ManuelSchneid3r:
sorry i cant merge this. i just merged a PR that claims to fix crashes and this commit changes a _lot_. who is going to maintain this? nobody wrote it. maybe if you can find some people that review it im fine with it, but I am not going to maintain it. should there ever be an issue with this code nobody feels responsible. maybe you could ping the author to take care of it.

Issue #215, opened 2025-04-20
Albert freezes when triggering Bitwarden Python plugin ('bw' query)


mg-dev25:
### Source

Open Build Service

### App logs

```Text
19:30:40 [debg:albert]        Albert version: 0.27.8
19:30:40 [debg:albert]            Build date: Oct 27 2017 00:00:00
19:30:40 [debg:albert]            Qt version: 6.8.2
19:30:40 [debg:albert]             Build ABI: x86_64-little_endian-lp64
19:30:40 [debg:albert]    Build architecture: x86_64
19:30:40 [debg:albert]      CPU architecture: x86_64
19:30:40 [debg:albert]           Kernel type: linux
19:30:40 [debg:albert]        Kernel version: 6.14.2-201.nobara.fc41.x86_64
19:30:40 [debg:albert]                    OS: Nobara Linux 41 (KDE Plasma)
19:30:40 [debg:albert]               OS type: nobara
19:30:40 [debg:albert]            OS version: 41
19:30:40 [debg:albert]         Platform name: wayland
19:30:40 [debg:albert]            Style name: breeze
19:30:40 [debg:albert]              Language: English
19:30:40 [debg:albert]                Locale: en_US
19:30:40 [debg:albert]          Ui languages: en-US, en-Latn-US, en
19:30:40 [debg:albert]       Binary location: /usr/bin/albert
19:30:40 [debg:albert]           Working dir: /home/mg
19:30:40 [debg:albert]             Arguments: albert
19:30:40 [debg:albert] ENVIRONMENT:
19:30:40 [debg:albert]         QT_LOGGING_RULES: albert*=true
19:30:40 [debg:albert]               PATH: /home/mg/.local/share/zinit/plugins/...:/sbin (trimmed for brevity)
19:30:40 [debg:albert]       WAYLAND_DISPLAY: wayland-0
19:30:40 [debg:albert]         PYTHON_ENV_INITIALIZED: 1
19:30:40 [debg:albert] ... (other environment variables) ...
19:30:40 [debg:albert] Loading translations
...
19:30:40 [info:albert] Searching native plugins in /usr/lib64/albert
...
19:30:40 [debg:albert] Found valid native plugin /usr/lib64/albert/python.so
...
19:30:40 [debg:albert] Using translations for 'python' from :/i18n/python_en.qm
19:30:40 [debg:albert] 3 ms spent loading plugin 'python'
19:30:40 [debg:albert.python] Python version: 3.13.0
[...]

PR #221, opened 2025-06-21
[bitwarden] Fix freeze when triggering bitwarden

MagneFire:
A freeze can be observed when triggering the bitwarden plugin. This seems to occur the second time `query.add` is called. To fix this issue the same logic as the `kill` plugin is used: Add all items to a local array before finally adding this using `query.add`

This fixes https://github.com/albertlauncher/python/issues/215


ManuelSchneid3r:
👍

PR #223, opened 2025-07-03
Fix syncthing plugin

ngencokamin:
Syncthing wasn't working properly. Was throwing errors about st not being defined any time the user types in albert, and the error `[crit:albert.python] Unable to cast Python instance of type <class 'NoneType'> to C++ type 'QString'` when opening settings, displaying a blank window instead of the api key field. This PR:

- Adds handling to make sure the api key is never None type, fixing the settings issue
- Adds handling for missing and invalid api keys
    - Shows distinct messages for each scenarios

**Screenshots**
image
Valid api key

image
Invalid api key

image
No api key

### DIFF
--- a/syncthing/__init__.py
+++ b/syncthing/__init__.py
@@ -41,7 +41,6 @@ class Plugin(PluginInstance, GlobalQueryHandler):
     @property
     def api_key(self) -> str:
         return self._api_key
-        # return '1234'
 
     @api_key.setter
     def api_key(self, value: str):


ngencokamin:
Oh forgot to include, here's a before and after of the settings menu
image
image

ngencokamin:
> @ngencokamin lets continue the discussion here. what have you changed to make the config widget work?

Hey sorry for the slow reply, been crazy busy with work. Lemme run through and break down my changes and why I made them/what they do

syncthing/__init__.py
```diff
@@ -22,14 +22,16 @@
 
 class Plugin(PluginInstance, GlobalQueryHandler):
 
-    config_key = 'syncthing_api_key'
+    config_key = 'api_key'
```
ngencokamin, line 25: This is to make the config key consistent with references throughout the rest of the script

syncthing/__init__.py
```diff
@@ -22,14 +22,16 @@
...
 
     def __init__(self):
         PluginInstance.__init__(self)
         GlobalQueryHandler.__init__(self)
 
         self.iconUrls = ["xdg:syncthing", f"file:{Path(__file__).parent}/syncthing.svg"]
         self._api_key = self.readConfig(self.config_key, str)
+        if self._api_key is None:
+            self._api_key = ''
```
ngencokamin, line 34: This fixes the widget not appearing with error `12:26:44 [crit:albert.python] Unable to cast Python instance of type <class 'NoneType'> to C++ type 'QString'`

Basically just initializes api_key to an empty string if an api key isn't found in the config. Otherwise it breaks because of it being `None`/`null`

syncthing/__init__.py
```diff
@@ -39,6 +41,7 @@ def defaultTrigger(self):
     @property
     def api_key(self) -> str:
         return self._api_key
+        # return '1234'
```
ngencokamin, line 44: I didn't mean to leave this in and will make a commit removing it

syncthing/__init__.py
```diff
@@ -65,76 +68,90 @@ def configWidget(self):
...
+                                    f"Shared with {folders_devices if folders_devices else 'nobody'}.",
+                            iconUrls=self.iconUrls,
+                            actions=[
+                                Action("scan", "Scan the folder",
+                                    lambda fid=folder_id: self.st.database.scan(fid)),
+                                Action("open", "Open this folder in file browser",
+                                    lambda p=f['path']: openFile(p))
+                            ]
+                        )
+                        results.append(RankItem(item, match))
```
ngencokamin, line 140: This bit `try` is meant to catch it when `syncthing` fails to initialize due to missing or invalid api key. Otherwise it spits errors with every character entered into the launcher. Plus, having it blank with no explanation outside of the console is not the most friendly design.

ngencokamin: I maybe could do the try except only around the bit that checks if `st` is initialized/valid? So that it's a little less general

syncthing/__init__.py
```diff
@@ -65,76 +68,90 @@ def configWidget(self):
...
-                                   lambda p=f['path']: openFile(p))
-                        ]
+                        id="invalid_key",
+                        text='Invalid API Key',
+                        subtext=f"API Key {self._api_key} is invalid. Please try entering your API key again in settings.",
+                        iconUrls=self.iconUrls
                     )
-                    results.append(RankItem(item, match))
-
+                results.append(RankItem(item, 0))
```
ngencokamin, line 156: So *in theory*, this except should only fire if the api key is invalid or missing. But I can see how we might want the logic to be a little less broad. Basically rn it just checks if the key is blank, at which point it reports missing, else it reports invalid.

ngencokamin (commented):
Here's a lil breakdown of the code choices I made! Sorry it took so long 😭

syncthing/__init__.py
ngencokamin: Okay, this is fixed now

ngencokamin:
> > @ngencokamin lets continue the discussion here. what have you changed to make the config widget work?
> 
> Hey sorry for the slow reply, been crazy busy with work. Lemme run through and break down my changes and why I made them/what they do



> @ngencokamin lets continue the discussion here. what have you changed to make the config widget work?

Oh I can also add better code comments if that'd be helpful

syncthing/__init__.py
Sharsie: Looking at the code, it seems that `self.st` is only defined when api key is set. So for anyone using this plugin and having that config present (maybe from a past version), this would be a breaking change. If I'm understanding this correctly, the plugin is not working properly for someone enabling it for the first time.
I would recommend keeping the old config key name not to force people into having to input it again.

syncthing/__init__.py
Sharsie: This could be shortened to 
```
self._api_key = self.readConfig(self.config_key, str) or ""
```

syncthing/__init__.py
Sharsie: I would suggest a few points to refactor this and make it more robust

First: I would look into `api_key` setter. It only sets `self.st` if the API key changes, but that includes if api key is empty. I'm not that familiar with Syncthing to know if this is allowed, but I assume you need an API key.
So instead of creating `self.st` every time, I would set it to `None` if api key is empty.

Second: You wouldn't need exception handling for empty API key. Instead, the first thing in `handleGlobalQuery` could be a check to `self.st` (that is now properly set due to the First point) and return the StandardItem with information about missing API key, something along the lines of

```
if not self.st:
  return [StandardItem(...)]
```

Third: You should keep the try block for any call to syncthing, do not wrap it only for the initial call. It could go down between the loop iterations and you need to catch those errors. However, printing out "Invalid API key" may not be appropriate, as Syncthing could just be down/not available for any reason. To make it more readable, exporting this logic into another function and then calling that function in a try except block is also an option

PR #225, opened 2025-07-07
[vscode_projects:1.11] Updates recents search to use latest vscode configuration, performance improvements

Sharsie:
This PR replaces searching through legacy `json` vscode configuration files by querying sqlite database. Recents are only stored and updated in the db in the latest versions.

It also improves documentation and introduces some performance optimizations and error handling.

Removes support for multiple paths per configuration and only uses a single path to read configuration from. Previously, this was implemented to support older versions of vscode, but sqlite has completely replaced this a while ago so I've cleaned up the code that deals with the multi-config feature.

It also enables searching through recents when only query trigger is supplied with empty query. Previously typing a query trigger yielded no results. I believe this improves UX by offering the latest entries immediately.

Comes bundled with https://github.com/albertlauncher/python/pull/224 - I honestly did not know how to version these properly without having to update one PR after the other gets merged. So ideally https://github.com/albertlauncher/python/pull/224 should be merged first.

### DIFF
--- a/vscode_projects/__init__.py
+++ b/vscode_projects/__init__.py
@@ -8,7 +8,7 @@ from dataclasses import dataclass
 from albert import *
 
 md_iid = "3.0"
-md_version = "1.9"
+md_version = "1.10"
 md_name = "VSCode projects"
 md_description = "Open VSCode projects"
 md_url = "https://github.com/albertlauncher/python/tree/master/vscode_projects"
@@ -103,7 +103,7 @@ class Plugin(PluginInstance, TriggerQueryHandler):
             warning(
                 "Project Manager search was enabled, but configuration file was not found")
             notif = Notification(
-                title=f"{self.name}",
+                title=self.name,
                 text=f"Configuration file was not found for the Project Manager extension. Please make sure the extension is installed."
             )
             notif.send()
@@ -359,7 +359,7 @@ Usecase with single VSCode instance - To reuse the VSCode window instead of open
             text=project.displayName,
             subtext=f"{subtext}{project.path}",
             iconUrls=self.iconUrls,
-            inputActionText=f"{query.trigger}{project.displayName}",
+            inputActionText=project.displayName,
             actions=actions,
         )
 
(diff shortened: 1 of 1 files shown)


Sharsie:
@Silureth I would appreciate if you could take a peek since it seems noone else would review the plugin.

Python is not my strong suit, but I've been using this code locally for months now. Though I made some improvements so it's presentable.

Sharsie:
Thanks for the feedback, yes that is likely, I do not use workspaces at all, so it may have different data structure. I intentionally skip single files, but that was more of a personal preference and I think workspace comes a single file <name>.code-workspace as it seems.

There was another option I exposed as a configuration value - custom command that can be used inside the project's workdir, which pretty much [changes dir to the workdir of the project](https://github.com/albertlauncher/python/blob/main/vscode_projects/__init__.py#L339) and then runs whatever command you set (e.g. `code .`) - I'm using it in combination with direnv to load the environment before opening vscode (e.g. `direnv exec . code .`)

I think it goes against the concept of workspaces, where there is no such thing as a single working directory, since it can contain multiple different system paths.

Definitely makes sense to support it, but I have to think about handling both of these use-cases - likely disabling custom commands for single-file recents, but that introduces a conditional behavior I don't like very much. I'll ponder what to do with that.

Silureth (approved):
Recents works as intended for folder based projects.

Sharsie:
https://github.com/albertlauncher/python/pull/225/commits/dc96d2ce75cd747df3b30d7f05434397a6e45fb7 expands support for workspaces and single file recent entries

ManuelSchneid3r:
Hi, I split the python plugin repository in submodules per plugin. please apply this PR against the repo albert-plugin-python-<project_name>. I'm sorry for the inconvenience.

Sharsie:
New MR is in https://github.com/albertlauncher/albert-plugin-python-vscode-projects/pull/1

I like the idea, I hope this is not going to kill PR reviews completely though.

